PR #42, opened 2019-08-29
Make json conversion more generic

### LINKED ISSUE #35: @@TRACES@@: MetricBatchJsonGenerator refactor to loosen Metric coupling
The MetricBatchJsonGenerator should be refactored so that it is less dependent on metrics alone.  At the end of this refactoring, we should have a class that can marshal any telemetry type (of which there are just metrics right now) to json.

When marshaling metrics, the resulting json should still confirm to the current format.

breedx-nr:
Resolves #35.

There was quite a bit of refactoring around the json generation.  Since things have been torn apart and moved around, it might make the most sense to go commit-by-commit.

A few items of note:
* The improper use of "Generator" is now removed from class names
* The base type of all telemetry is now `Telemetry` (interface)
* `MetricBatch` is a type-specific sub of `TelemetryBatch<Metric>`
* Json generation is done thru `TelemetryBatchJson` which is no longer coupled to `Metric`


telemetry_components/src/main/java/com/newrelic/telemetry/GsonMetricToJson.java
```diff
@@ -18,16 +19,16 @@
  * manually are to decouple the SDK types (e.g., Count, Gauge, Summary) from the API structure, and
  * to avoid excess object creation and allocation.
  */
-public class MetricGsonGenerator implements MetricJsonGenerator {
+public class GsonMetricToJson implements MetricToJson {
```
jasonjkeller: Why not call the implementation `MetricToGson`?

PR #50, opened 2019-09-04
Add implementation and test for attributes copy

### LINKED ISSUE #47: Add the ability to copy Attributes
There are a few cases in the sdk (and by the exporter) where having the ability to copy an `Attributes` instance would be helpful.  We should add a method like `Attributes.copy()` that returns a new instance backed by a new map.  Mutations to the resulting copied instance should *not* mutate the original instance.

jeemar:
Hi all, please take a look. Hopefully this resolves #47. Kind wrote this quickly, might be bugs...

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
@@ -25,6 +25,16 @@ import lombok.ToString;
 public class Attributes {
   private final Map<String, Object> rawAttributes = new HashMap<>();
 
+  public Attributes() {}
+
+  public Attributes(Attributes other) {
+    other.rawAttributes.forEach(rawAttributes::put);
+  }
+
+  public Attributes copy() {
+    return new Attributes(this);
+  }
+
   /**
    * Add a string-valued attribute.
    *
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
@@ -25,12 +25,15 @@ import lombok.ToString;
 public class Attributes {
   private final Map<String, Object> rawAttributes = new HashMap<>();
 
+  /** Creates an empty object */
   public Attributes() {}
 
-  public Attributes(Attributes other) {
-    other.rawAttributes.forEach(rawAttributes::put);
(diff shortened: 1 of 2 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
```diff
@@ -25,6 +25,16 @@
 public class Attributes {
   private final Map<String, Object> rawAttributes = new HashMap<>();
 
+  public Attributes() {}
+
+  public Attributes(Attributes other) {
```
XiXiaPdx: I think we might want this constructor to only be private. Let's verify.

XiXiaPdx: Thoughts on "original" instead of "other?

telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
```diff
@@ -25,6 +25,16 @@
...
+    other.rawAttributes.forEach(rawAttributes::put);
```
jkwatson: I think we can say `rawAttributes.putAll(other.rawAttributes)`

telemetry_core/src/main/java/com/newrelic/telemetry/Attributes.java
jkwatson: we could just call it `attributes`

jeemar: I'm okay with "original".

jkwatson:
Just add some javadoc to those new methods, and this should be good to 🚢

telemetry_core/src/test/java/com/newrelic/telemetry/AttributesTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Test
+  void testCopyMethod() {
+    Attributes a = new Attributes().put("superFoo", "openBar");
+    Attributes copyOfA = a.copy();
+
+    a.put("hello", "goodbye");
+    Assertions.assertTrue(a.asMap().containsKey("superFoo"));
+    Assertions.assertTrue(a.asMap().containsKey("hello"));
+    Assertions.assertFalse(copyOfA.asMap().containsKey("hello"));
```
jkwatson: let's static-import these methods, to clean up the code.

PR #55, opened 2019-09-05
Add copy method to Attributes

### LINKED ISSUE #47: Add the ability to copy Attributes

breedx-nr:
Resolves #47


breedx-nr:
Doh.

PR #64, opened 2019-09-05
Span integration test

### LINKED ISSUE #54: Create integration test(s) for the Trace/Span API

XiXiaPdx:
resolves #54

### DIFF
--- a/integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
+++ b/integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
@@ -15,7 +15,6 @@ import com.newrelic.telemetry.spans.SpanBatchSender;
 import java.net.URI;
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
@@ -34,7 +33,7 @@ import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
 import org.testcontainers.junit.jupiter.Testcontainers;
 
 @Testcontainers
-public class SpanApiIntegrationTest {
+class SpanApiIntegrationTest {
 
   private static final int SERVICE_PORT = 1080 + new Random().nextInt(900);
   private static String containerIpAddress;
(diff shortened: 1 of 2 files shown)


telemetry_examples/src/main/java/com/newrelic/telemetry/examples/SpanExample.java
```diff
@@ -49,6 +50,7 @@ public static void main(String[] args) throws Exception {
             .httpPoster(new OkHttpPoster(Duration.ofSeconds(5)))
             .attributesJson(new AttributesGson(new Gson()))
             .enableAuditLogging()
+            .uriOverride(URI.create("https://staging-trace-api.newrelic.com"))
```
breedx-nr: I'm thinking that we don't want our examples pointing at staging, we should just omit this for clarity and brevity.

integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+                .withMethod("POST")
+                .withPath("/trace/v1")
+                .withBody(
+                    json(
+                        new SpanPayload[] {expectedPayload},
+                        MediaType.JSON_UTF_8,
+                        MatchType.ONLY_MATCHING_FIELDS))
+                .withHeader("User-Agent", "NewRelic-Java-TelemetrySDK/.*")
+                .withHeader("Content-Type", "application/json; charset=utf-8")
+                .withHeader("Content-Length", ".*"))
```
breedx-nr: Is this helpful, or can we omit the wildcard length check?

integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+            ImmutableMap.of("trace.id", "123456", "attributes", singletonMap("key1", "val1")),
+            singletonList(
+                ImmutableMap.<String, Object>builder()
+                    .put("id", "6f377f46-3f10-4e8e-8390-7717044cdbe9")
+                    .put("trace.id", "123456")
+                    .put("timestamp", 55555)
+                    .put(
+                        "attributes",
+                        ImmutableMap.of(
+                            "duration.ms",
```
breedx-nr: nitpick: might be nice to see if these pairs will fit on the same line (reads more like mappings that way).

integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
jkwatson: I think this is a good thing to check...it's well-behaved to provide a content length header.

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/SpanExample.java
jkwatson: 👍

integration_test/src/test/java/com/newrelic/telemetry/SpanApiIntegrationTest.java
XiXiaPdx: fixed

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/SpanExample.java
XiXiaPdx: removed

Issue #58, opened 2019-09-05
Create a "simple" factory/builder for the SpanBatchSender


jkwatson:
The analogue of the SimpleMetricBatchSender

jkwatson:
this has been done for quite a while. closing.

PR #66, opened 2019-09-06
create a SimpleSpanBatchSender and wire up to the example

breedx-nr:
This addresses #58.

A few other notable changes:
* adds builder verification for the attributes json instance being supplied (wasn't checked before)
* adds test coverage around the builder verifications

I'm also looking for feedback on the fact that I exposed 4 methods on the `SimpleSpanBatchSender` -- 2 that return the builder in case the user wants additional customization (like in the example) and 2 that return the 'Sender directly.

### DIFF
--- a/telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
+++ b/telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
@@ -1,7 +1,6 @@
 package com.newrelic.telemetry;
 
 import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
 import com.newrelic.telemetry.http.HttpPoster;
 import com.newrelic.telemetry.spans.SpanBatchSender;
 import com.newrelic.telemetry.spans.SpanBatchSenderBuilder;
@@ -23,10 +22,9 @@ public class SimpleSpanBatchSender {
 
   public static SpanBatchSenderBuilder builder(String apiKey, Duration callTimeout) {
     HttpPoster http = new OkHttpPoster(callTimeout);
-    Gson gson = new GsonBuilder().create();
     return SpanBatchSender.builder()
         .apiKey(apiKey)
         .httpPoster(http)
-        .attributesJson(new AttributesGson(gson));
+        .attributesJson(new AttributesGson(new Gson()));
   }
 }
--- a/telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
+++ b/telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
@@ -1,0 +1,32 @@
+package com.newrelic.telemetry;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.newrelic.telemetry.http.HttpPoster;
+import com.newrelic.telemetry.spans.SpanBatchSender;
+import com.newrelic.telemetry.spans.SpanBatchSenderBuilder;
+import java.time.Duration;
+
+public class SimpleSpanBatchSender {
+
+  public static SpanBatchSender build(String apiKey) {
+    return build(apiKey, Duration.ZERO);
+  }
+
+  public static SpanBatchSenderBuilder builder(String apiKey) {
+    return builder(apiKey);
+  }
+
+  public static SpanBatchSender build(String apiKey, Duration callTimeout) {
+    return builder(apiKey, callTimeout).build();
+  }
+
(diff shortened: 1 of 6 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSenderBuilder.java
```diff
@@ -35,24 +35,28 @@ public MetricBatchSender build() {
...
     Utils.verifyNonNull(attributesJson, "an AttributesJson implementation is required.");
 
-    if (metricsUrl == null) {
-      try {
-        metricsUrl = constructMetricsUrlWithHost(URI.create("https://metric-api.newrelic.com/"));
-      } catch (MalformedURLException e) {
-        throw new UncheckedIOException("Bad hardcoded URL", e);
-      }
-    }
+    URL url = getOrDefaultMetricsUrl();
```
jkwatson: I like it

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public static SpanBatchSenderBuilder builder(String apiKey) {
+    return builder(apiKey);
+  }
+
+  public static SpanBatchSender build(String apiKey, Duration callTimeout) {
+    return builder(apiKey, callTimeout).build();
+  }
+
+  public static SpanBatchSenderBuilder builder(String apiKey, Duration callTimeout) {
```
jkwatson: I'd like to change the metric version of this to look more like this, as well.

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.gson.GsonBuilder;
+import com.newrelic.telemetry.http.HttpPoster;
+import com.newrelic.telemetry.spans.SpanBatchSender;
+import com.newrelic.telemetry.spans.SpanBatchSenderBuilder;
+import java.time.Duration;
+
+public class SimpleSpanBatchSender {
+
+  public static SpanBatchSender build(String apiKey) {
+    return build(apiKey, Duration.ZERO);
```
jkwatson: I don't think ZERO is a good value here. We should choose a good default. ZERO means infinite timeout, and we've gotten feedback that this is a bad choice.

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
```diff
@@ -0,0 +1,32 @@
...
+    return builder(apiKey);
+  }
+
+  public static SpanBatchSender build(String apiKey, Duration callTimeout) {
+    return builder(apiKey, callTimeout).build();
+  }
+
+  public static SpanBatchSenderBuilder builder(String apiKey, Duration callTimeout) {
+    HttpPoster http = new OkHttpPoster(callTimeout);
+    Gson gson = new GsonBuilder().create();
```
jkwatson: does this do anything different than just `new Gson()` ?

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
breedx-nr: Heh, yeah, I was borrowing heavily from the other simple sender.  I agree tho and will change it to 2 seconds as a reasonable starting point.

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
breedx-nr: The docs say
```
 * <p>Use this builder to construct a {@link Gson} instance when you need to set configuration
 * options other than the default. For {@link Gson} with default configuration, it is simpler to
 * use {@code new Gson()}. {@code GsonBuilder} is best used by creating it, and then invoking its
 * various configuration methods, and finally calling create.</p>
 *
```
so yeah, I'll change it to the simpler constructor.  Good catch!  🏆

telemetry_components/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
breedx-nr: I picked that up in this PR now too.

breedx-nr:
Ready for another looksie.

jkwatson:
let's ship it. you need to rebase, and we need to write about a million lines of javadoc.

PR #67, opened 2019-09-06
remove GSON and replace with a piece of the gson codebase

jkwatson:
this has been properly attributed via a third party notices file


telemetry_core/src/main/java/com/newrelic/telemetry/json/AttributesJson.java
```diff
@@ -1,36 +1,50 @@
...
+import java.io.IOException;
+import java.io.StringWriter;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
-public interface AttributesJson {
+public class AttributesJson {
 
-  /**
```
breedx-nr: Oof, isn't it still helpful to have this javadoc here?

telemetry_core/src/main/java/com/newrelic/telemetry/json/AttributesJson.java
```diff
@@ -1,36 +1,50 @@
...
-   * Call this to remove illegal (NaN, -Infinity, null, etc) values from the attributes before
-   * generating JSON for them.
-   */
-  default Map<String, Object> filterIllegalValues(Map<String, Object> attributes) {
+      for (Map.Entry<String, Object> attribute : filteredAttributes.entrySet()) {
+        Object value = attribute.getValue();
+
+        if (value instanceof Boolean) {
+          jsonWriter.name(attribute.getKey()).value((boolean) value);
```
breedx-nr: the `jsonWriter.name(attribute.getKey())` bit is repeated in all cases.  We should pull it up.

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricToJson.java
```diff
@@ -1,100 +1,93 @@
...
-   *   "timestamp": 1560807066910,
-   *   "interval.ms": 5000,
-   *   "attributes": {
-   *     "key1": "value1",
-   *     "key2", 555,
-   *     "key3", false
-   *   }
-   * }
-   * </pre>
-   */
```
breedx-nr: Same question about this javadoc -- shouldn't we keep it for helpful clarity?

telemetry_core/src/main/java/com/newrelic/telemetry/json/AttributesJson.java
jkwatson: I don't think anyone outside of us will care about it, because users no longer have to provide it, so no.

telemetry_core/src/main/java/com/newrelic/telemetry/json/AttributesJson.java
jkwatson: this wasn't changed from before. we can clean this up after this PR, I guess?

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricToJson.java
jkwatson: It would only be for us, since no one ever needs to write a json thingee any more (except us)

Issue #70, opened 2019-09-06
Add Javadocs for Spans 


XiXiaPdx:
Create all necessary Javadocs for Span related sources

PR #75, opened 2019-09-06
Add javadoc to public facing methods to some span classes

jeemar:
Addresses #70

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/spans/SpanBatchSenderBuilder.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/spans/SpanBatchSenderBuilder.java
@@ -21,6 +21,7 @@ import java.net.URL;
 
 public class SpanBatchSenderBuilder {
   private static final String spansPath = "/trace/v1";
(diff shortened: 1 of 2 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/spans/SpanBatchSenderBuilder.java
```diff
@@ -21,6 +21,7 @@
 
 public class SpanBatchSenderBuilder {
   private static final String spansPath = "/trace/v1";
+  private static final String DEFAULT_TRACE_URL = "https://trace-api.newrelic.com/";
```
jeemar: I pulled this thing out so that I could reference it in a javadoc without having to re-type it and risk this the field and doc getting out of sync.

jkwatson: 👍

breedx-nr: Solid idea!

Issue #65, opened 2019-09-06
Verify that `version.id` is populated in the user agent for spans


breedx-nr:
The version should be obtained from the package of the BatchDataSender inside the jar.  We should confirm that this value is being fetched correctly and put into the user agent when using the jar (and not just running from the IDE).

breedx-nr:
We verified that this is broken due to it being missing from the jar manifest.

breedx-nr:
Confirmed that as of 0.3.1-SNAPSHOT that this works and the user agent is configured correctly.

PR #84, opened 2019-09-06
Fix jar manifests

breedx-nr:


gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Here is where we manage the version
-releaseVersion=0.3.1-SNAPSHOT
+releaseVersion=0.3.1
```
jkwatson: this scares me because we haven't actually tested that this reports correctly

breedx-nr: This is only intended to fix up the manifest, shouldn't have any functional changes (except it might also fix #65 )

PR #87, opened 2019-09-09
fix a infinite recursion in the simple span batch sender

### LINKED ISSUE #85: Stack overflow when trying to use SimpleSpanBatchSender.builder(String apiKey)
it just calls itself.  🤦‍♂

jkwatson:
resolves #85


telemetry-http-okhttp/src/main/java/com/newrelic/telemetry/SimpleSpanBatchSender.java
```diff
@@ -34,7 +34,7 @@ public static SpanBatchSender build(String apiKey) {
    *     Relic API Keys</a>
    */
   public static SpanBatchSenderBuilder builder(String apiKey) {
-    return builder(apiKey);
```
breedx-nr: Yeah that'll do it.

PR #89, opened 2019-09-09
Add error to spans

### LINKED ISSUE #86: Add the error attribute as a first-class option for Spans
If you set an the attribute `error` to `true` in a span, then the DT UI knows it was an span with an error in it. we should provide that as a first-class option in our Span builder.

breedx-nr:
This resolves #86 .

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
@@ -150,7 +150,12 @@ public final class Span implements Telemetry {
       return this;
     }
 
-    public SpanBuilder isError() {
+    /**
+     * Call this to indicate that the span contains an error condition.
+     *
+     * @return The SpanBuilder instance with the error field set to true
+     */
+    public SpanBuilder withError() {
       this.error = true;
       return this;
     }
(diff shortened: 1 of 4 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
```diff
@@ -146,9 +150,15 @@ public SpanBuilder parentId(String parentId) {
       return this;
     }
 
+    public SpanBuilder isError() {
```
jkwatson: javadoc, por favor!

telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
```diff
@@ -146,9 +150,14 @@ public SpanBuilder parentId(String parentId) {
       return this;
     }
 
+    public SpanBuilder isError(){
```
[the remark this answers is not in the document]
breedx-nr: I like it, I think the name isn't great as it is now.  Will change!

telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
breedx-nr: Yes.

breedx-nr:
Ready for another  👁 👁 ···

PR #90, opened 2019-09-09
clean up the READMEs a bit and add spans to the TelemetryClientExample

### LINKED ISSUE #88: Add a section to the README for getting started as a user of the SDK
We have one for developers, but not for users.

jkwatson:
resolves #88

### DIFF
--- a/telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
+++ b/telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
@@ -51,7 +51,7 @@ public class TelemetryClientExample {
     commonAttributes.put("appName", "testApplication");
     commonAttributes.put("environment", "staging");
 
-    sendSampleSpan(spanBatchSender, commonAttributes);
+    sendSampleSpan(telemetryClient, commonAttributes);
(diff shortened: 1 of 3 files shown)


telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
```diff
@@ -67,13 +85,24 @@ public static void main(String[] args) throws Exception {
...
+        Span.builder(UUID.randomUUID().toString())
+            .timestamp(System.currentTimeMillis())
+            .durationMs(150d)
+            .serviceName("Test Service")
+            .name("testSpan")
+            .build();
+    String traceId = UUID.randomUUID().toString();
+    SpanBatch spanBatch =
+        new SpanBatch(Collections.singleton(sampleSpan), commonAttributes, traceId);
+    spanBatchSender.sendBatch(spanBatch);
```
breedx-nr: I think this example should call through the `TelemetryClient instance` instead of the sender, right?

jkwatson: oh mos def. let me fix

PR #97, opened 2019-09-11
add SendBatch to handle multiple batches

### LINKED ISSUE #5: Add the ability to send a batch of metric batches
The backend API supports sending multiple batches in one payload. We should support that in the SDK.

(Migrate to Jira)

XiXiaPdx:
taking a stab at sending multiple batches...am I heading in the right direction?


jkwatson:
The underlying API (s) support sending multiple batches in a single request. I think we should try to do that, rather than just loop and send multiple batches. The split logic will be impacted, as well as the json generation.

PR #103, opened 2019-09-20
Deprecate the redundant traceUrl method.

jkwatson:


telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
```diff
@@ -11,7 +11,7 @@
 
 /**
  * Spans are New Relic's analog of OpenTracing spans. They can represent events such as external
- * calls individual operations, and datastore calls.
+ * calls, individual operations, and datastore calls.
```
breedx-nr: weirdest oxford comma ever!

PR #104, opened 2019-10-01
Attributes should be used when fields are missing

### LINKED ISSUE #102: If a span has standard values set in attributes, but not in the Span object, the marshaller overrides them.
For example, if you set `service.name` as a attribute on a span, but don't explicitly set the `serviceName` field via the `SpanBuilder`, then the resulting `service.name` attribute is not written at all.

breedx-nr:
This resolves #102

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/spans/json/SpanJsonTelemetryBlockWriter.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/spans/json/SpanJsonTelemetryBlockWriter.java
@@ -35,7 +35,7 @@ public class SpanJsonTelemetryBlockWriter {
           jsonWriter.name("trace.id").value(span.getTraceId());
         }
         jsonWriter.name("timestamp").value(span.getTimestamp());
-        Map<String, Object> enhancedAttributes = new AttributesEnhancer(span).apply();
+        Map<String, Object> enhancedAttributes = new AttributesEnhancer(span).enhance();
         jsonWriter.name("attributes").jsonValue(attributesJson.toJson(enhancedAttributes));
         jsonWriter.endObject();
       }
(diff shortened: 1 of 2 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/spans/json/SpanJsonTelemetryBlockWriter.java
```diff
@@ -40,15 +45,34 @@ public void appendTelemetryJson(SpanBatch batch, JsonWriter jsonWriter) {
...
+
+    private final Span span;
+    private final Map<String, Object> newAttrs;
+
+    AttributesEnhancer(Span span) {
+      this.span = span;
+      this.newAttrs = new HashMap<>(span.getAttributes().asMap());
+    }
+
+    Map<String, Object> apply() {
```
jkwatson: why not call this `enhance`?

breedx-nr: Into it.

jkwatson:
much easier to just use `putIfAbsent` for this, I think.

breedx-nr:
Agreed.  Totally derped on the existence of that.

telemetry_core/src/main/java/com/newrelic/telemetry/spans/json/SpanJsonTelemetryBlockWriter.java
```diff
@@ -42,10 +42,10 @@ public void appendTelemetryJson(SpanBatch batch, JsonWriter jsonWriter) {
...
   private Map<String, Object> enhanceAttributes(Span span) {
     Map<String, Object> result = new HashMap<>(span.getAttributes().asMap());
-    result.put("name", span.getName());
-    result.put("parent.id", span.getParentId());
-    result.put("duration.ms", span.getDurationMs());
-    result.put("service.name", span.getServiceName());
+    result.putIfAbsent("name", span.getName());
+    result.putIfAbsent("parent.id", span.getParentId());
+    result.putIfAbsent("duration.ms", span.getDurationMs());
+    result.putIfAbsent("service.name", span.getServiceName());
```
jkwatson: love it!

PR #110, opened 2019-12-05
Allow Summary metrics to have null min or max values

kscaldef:
Not all metric sources provide both min and max values for all metrics which we might
want to map to Summaries.  This allows either of these fields to be null, which the
NewRelic data backend will interpret as a double NaN and treat appropriately.


jkwatson:
This looks like a great change. Can you add some quick unit tests for these cases?

jkwatson:
@kscaldef looks like unit tests are failing for this, as well

kscaldef:
@jkwatson updated that test to reflect the change.

jkwatson:
@kscaldef is there urgency on getting this change into an official release? If so, I'll go ahead and kick off the release process.

PR #113, opened 2019-12-26
Enable providing a custom user agent

### LINKED ISSUE #109: Enable custom user agent suffixes for exporters
See https://github.com/newrelic/newrelic-telemetry-sdk-specs/blob/master/communication.md#user-agent

jkwatson:
resolves #109

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSenderBuilder.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSenderBuilder.java
@@ -25,6 +25,7 @@ public class MetricBatchSenderBuilder {
   private HttpPoster httpPoster;
   private URL metricsUrl;
   private boolean auditLoggingEnabled = false;
+  private String additionalUserAgent;
 
   /**
    * Build the final {@link MetricBatchSender}.
@@ -41,7 +42,8 @@ public class MetricBatchSenderBuilder {
         new MetricBatchMarshaller(
             new MetricBatchJsonCommonBlockWriter(new AttributesJson()),
             new MetricBatchJsonTelemetryBlockWriter(new MetricToJson()));
-    BatchDataSender sender = new BatchDataSender(httpPoster, apiKey, url, auditLoggingEnabled);
+    BatchDataSender sender =
+        new BatchDataSender(httpPoster, apiKey, url, auditLoggingEnabled, additionalUserAgent);
 
     return new MetricBatchSender(marshaller, sender);
   }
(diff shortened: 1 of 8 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSenderBuilder.java
```diff
@@ -80,16 +82,41 @@ public MetricBatchSenderBuilder enableAuditLogging() {
...
+   * non-empty. The version is optional, although highly recommended.
+   */
+  public MetricBatchSenderBuilder additionalUserAgent(String product, String version) {
+    Utils.verifyNonNull(product, "Product cannot be null in the additional user-agent.");
+    if (version == null || version.isEmpty()) {
+      additionalUserAgent = product;
+    } else {
+      additionalUserAgent = product + "/" + version;
+    }
+    return this;
```
jeffalder: I understand why you chose `additional`, but it can't be called multiple times. I might think that the additive nature of the name means it would keep appending. Maybe "secondUserAgent" or "secondaryUserAgent"?

telemetry_core/src/main/java/com/newrelic/telemetry/spans/SpanBatchSenderBuilder.java
```diff
@@ -110,6 +114,20 @@ public SpanBatchSenderBuilder traceUrl(URL traceUrl) {
     return this;
   }
 
+  /**
+   * Provide additional user agent information. The product is required to be non-null and
+   * non-empty. The version is optional, although highly recommended.
+   */
+  public SpanBatchSenderBuilder additionalUserAgent(String product, String version) {
+    Utils.verifyNonNull(product, "Product cannot be null in the additional user-agent.");
```
jeffalder: What if `product` is an empty string? `verifyNonNull` only checks nullness.

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSenderBuilder.java
jkwatson: I like `secondaryUserAgent`. Will make it so.

telemetry_core/src/main/java/com/newrelic/telemetry/spans/SpanBatchSenderBuilder.java
jkwatson: good call. I'll add a method to `Utils` to check for null or empty Strings.

jkwatson:
@jeffalder updated based on your excellent comments. Also, added a couple of unit tests.

PR #121, opened 2020-02-26
Log filtered count

### LINKED ISSUE #107: Log the number of metrics dropped per batch
When we're dropping metrics due to invalid values, it would be good to log (at DEBUG?) the number of metrics that we're dropping.

breedx-nr:
This fixes #107

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
@@ -4,16 +4,15 @@
  */
 package com.newrelic.telemetry.metrics.json;
 
-import com.newrelic.telemetry.metrics.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static java.lang.Double.isFinite;
 
+import com.newrelic.telemetry.metrics.*;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-
-import static java.lang.Double.isFinite;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class MetricBatchJsonTelemetryBlockWriter {
 
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
@@ -4,15 +4,16 @@
  */
 package com.newrelic.telemetry.metrics.json;
 
-import static java.lang.Double.isFinite;
-
 import com.newrelic.telemetry.metrics.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
```diff
@@ -27,13 +29,21 @@ public void appendTelemetryJson(MetricBatch batch, StringBuilder builder) {
...
+    AtomicInteger filteredCount = new AtomicInteger();
     builder.append(
         metrics
             .stream()
             .filter(this::isValid)
             .map(this::toJsonString)
+            .peek(x -> filteredCount.getAndIncrement())
             .collect(Collectors.joining(",")));
 
+    if (filteredCount.get() > 0) {
```
jkwatson: this looks like we're counting the number that got retained, rather than filtered out. am I crazy?

breedx-nr: Derp, you're right.  Lemme apply mathematics.

jeffalder: I mean, I assume someone tested it....

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
```diff
@@ -27,13 +28,21 @@ public void appendTelemetryJson(MetricBatch batch, StringBuilder builder) {
     builder.append("\"metrics\":").append("[");
     Collection<Metric> metrics = batch.getTelemetry();
 
+    AtomicInteger filteredCount = new AtomicInteger();
```
jkwatson: I'd rename this to "retainedCount"

breedx-nr: Sounds like a good change to me.

breedx-nr:
I kinda want to log these at info.  If it's happening a lot, users should know that data is being dropped.

jkwatson:
I think this was originally requested by Kevin S, maybe? Worth reaching out about how noisy he thinks he'd like it.

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/json/MetricBatchJsonTelemetryBlockWriter.java
```diff
@@ -38,7 +38,7 @@ public void appendTelemetryJson(MetricBatch batch, StringBuilder builder) {
             .collect(Collectors.joining(",")));
 
     if (retainedCount.get() != metrics.size()) {
-      logger.debug(
+      logger.info(
```
jkwatson: my only concern with "you should fix this" is that if the "bad" data is coming from exporters, or from libraries (like micrometer), this will end up being 100% non-actionable by the end-user, and they'll just have to turn off all logging at INFO to get rid of the noise. Not saying it's wrong...just something to consider.

breedx-nr: Fair enough.  I think I'm erring on the side of making it visible by default so that newly developed code doesn't introduce a hidden problem.  I totally hear your concern tho.

Issue #108, opened 2019-11-04
Implement the retry logic from the spec


jkwatson:
See here for details:  https://github.com/newrelic/newrelic-telemetry-sdk-specs/blob/master/communication.md#graceful-degradation

breedx-nr:
Resolved by #124...though there was talk about making the strategy pluggable some day.

jkwatson:
Yep. you created another issue for that. :). #126

PR #123, opened 2020-02-27
Add metric batch builder

breedx-nr:
This adds a convenience builder to `MetricBatch` and `MetricBuffer` so that service name and instrumentation provider can be emphasized.  Hopefully this resolves #101 .

PR #124, opened 2020-02-28
Retry backoff

breedx-nr:
This addresses #108

### DIFF
--- a/telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
@@ -1,0 +1,39 @@
+package com.newrelic.telemetry;
+
+public class Backoff {
+
+    private final long maxBackoffTimeMs;
+    private final long backoffFactorMs;
+    private final int maxRetries;
+    private int numRetries = 0;
+
+    public Backoff(long maxBackoffTimeMs, long backoffFactorMs, int maxRetries) {
+        this.maxBackoffTimeMs = maxBackoffTimeMs;
+        this.backoffFactorMs = backoffFactorMs;
+        this.maxRetries = maxRetries;
+    }
+
+    public static Backoff defaultBackoff() {
+        return new Backoff(15000, 1000, 10);
+    }
+
+    /**
+     * Computes the next wait time and does NOT actually sleep.  This increases
+     * the tracked number of retries;
+     *
+     * @return the next amount of time to wait, or -1 if max retries exceeded.
+     */
+    public long nextWaitMs() {
+        numRetries++;
+        int n = numRetries - 1;
+
+        if(n == 0){
+            return 0;
+        }
+        if(n >= maxRetries){
+            return -1;
+        }
+        return (long) Math.min(maxBackoffTimeMs, backoffFactorMs * Math.pow(2, n - 1));
+    }
+
+}
--- a/telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
@@ -2,38 +2,37 @@ package com.newrelic.telemetry;
 
 public class Backoff {
 
-    private final long maxBackoffTimeMs;
-    private final long backoffFactorMs;
-    private final int maxRetries;
-    private int numRetries = 0;
+  private final long maxBackoffTimeMs;
+  private final long backoffFactorMs;
+  private final int maxRetries;
+  private int numRetries = 0;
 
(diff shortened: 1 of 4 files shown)


breedx-nr:
This builds on top of #123 and will be rebased.

telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
```diff
@@ -0,0 +1,38 @@
...
+  private int numRetries = 0;
+
+  public Backoff(long maxBackoffTimeMs, long backoffFactorMs, int maxRetries) {
+    this.maxBackoffTimeMs = maxBackoffTimeMs;
+    this.backoffFactorMs = backoffFactorMs;
+    this.maxRetries = maxRetries;
+  }
+
+  public static Backoff defaultBackoff() {
+    return new Backoff(15000, 1000, 10);
```
jkwatson: whoa. this is hard to read. I wonder if we want durations or something to make this clearer?

jkwatson: or maybe some named parameters.  ;)

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -118,19 +124,11 @@ private void retry(
...
-      timeUnit = TimeUnit.SECONDS;
-    } else {
-      newWaitTime = preWaitTime * 2;
-    }
-    LOG.info("Metric batch sending failed. Backing off {} {}", newWaitTime, timeUnit);
-    scheduleBatchSend(sender, batch, newWaitTime, timeUnit);
+      BatchSender sender, TelemetryBatch<? extends Telemetry> batch, Backoff backoff) {
+
+    long newWaitTime = backoff.nextWaitMs();
+    LOG.info("Metric batch sending failed. Backing off {} {}", newWaitTime, TimeUnit.MILLISECONDS);
```
jkwatson: I think you're missing the handling of the -1 case here.

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -71,26 +71,31 @@ public void sendBatch(SpanBatch batch) {
   private void scheduleBatchSend(
       BatchSender sender,
       TelemetryBatch<? extends Telemetry> batch,
-      int waitTime,
+      long waitTime,
       TimeUnit timeUnit) {
-    executor.schedule(
-        () -> sendWithErrorHandling(sender, batch, waitTime, timeUnit), waitTime, timeUnit);
+    scheduleBatchSend(sender, batch, waitTime, timeUnit, Backoff.defaultBackoff());
```
jkwatson: are user-configurable backoff strategies planned for a later PR?

telemetry/src/main/java/com/newrelic/telemetry/Backoff.java
breedx-nr: I made a builder that is as close as we can get to named params in this stuff.  :)

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
breedx-nr: I don't follow what you mean by this.

breedx-nr: Oh NM, I see what you mean now and I think you're right.

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
breedx-nr: there's probably room for interpretation, but I didn't see in the spec that user configured times were required.  In any case, yes, prefer to do that in a future PR.

jkwatson:
```
For a backoff factor of 1 second, and a backoff max of 16 seconds, the retry delay interval should follow a pattern of [0, 1, 2, 4, 8, 16, 16, ...]. Subsequent retries should wait 16 seconds until the request has been retried the configured max retries number of times.

The total retry duration can be computed from the combination of backoff factor and backoff max. SDKs may provide a function to configure retry behavior by specifying the total retry duration instead of max retries.
```

It certainly mentions "configure" a bunch, but a bit obliquely. We should PR into the specs to make it very clear what should be user-configurable and what the defaults should be.

breedx-nr:
Created this issue: https://github.com/newrelic/newrelic-telemetry-sdk-java/issues/125

breedx-nr:
Also made #126.

PR #130, opened 2020-03-11
clean up the example to use more canonical attribute names

jkwatson:


telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
```diff
@@ -39,8 +38,7 @@ public static void main(String[] args) throws Exception {
     TelemetryClient telemetryClient = new TelemetryClient(batchSender, spanBatchSender);
 
     Attributes commonAttributes = new Attributes().put("exampleName", "TelemetryClientExample");
-    commonAttributes.put("host", InetAddress.getLocalHost().getHostName());
-    commonAttributes.put("appName", "testApplication");
+    commonAttributes.put("host.hostname", InetAddress.getLocalHost().getHostName());
```
breedx-nr: I like/grok the `host.hostname` change here, but where did `appName` run off to?  I think it should be `app.name` right?

jkwatson: it shouldn't. this should be `service.name` and is done below. It makes me think that we should make the `Attributes` smarter so people can do the right thing more easily at the batch level.

breedx-nr: Cool.

PR #131, opened 2020-03-19
'find and use data' section from docs team

zachelwood:
This is from Zach Elwood on NR docs team. 

This PR is to add a ‘how to find data’ section so that customers can find the data they send. I’ve included what I think are the applicable data types (spans and metrics) but let me know if there are other data types reported by this exporter. If you think this exporter will change to send other types of data in future, keep in mind it will be important to update this repo with link to that data type.

Also changed some instances of title case to sentence case; small thing but we want style to be sentence case and not title case, so just something to keep in mind when creating new content. Let me know if you think I changed sentence case stuff that shouldn't have been changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Why is this cool?
 
 Dimensional metrics and spans in New Relic! No agent required. 
 
-The telemetry SDK tries to be helpful, so your job of sending telemetry data to New Relic can be done in the right way, easily. We've covered all of the basics for you so you can focus on writing feature code directly related to your business need or interest.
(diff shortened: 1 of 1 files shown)


zachelwood:
realized we should leave 'Telemetry APIs' as that and not 'Telemetry SDKs' like I changed it to, at bottom. Sorry.

jkwatson:
feel free to update the PR to revert that change, @zachelwood

jeffalder:
@zachelwood  would you update your PR per your comment above? Then we can merge it in.

PR #132, opened 2020-03-25
Java11 native provider

kittylyst:


jeffalder:
@kittylyst I love this, Thanks! It's great to have an updated http client. It'll simplify the JFR reporter, for sure.

There are no updates to telemetry_examples; that would be useful given the new `provider` method.

What lead you to make `provider` static? I wouldn't expect a Sender to be created multiple times; I think it should generally be created once and reused.

It looks like the PR build failed; are you able to see the results? If so, can you fix up the PR to fix the errors?

kittylyst:
Thanks Jeff. The issue with the PR is that the integration_test target is broken out of the box on my machine, and rather than trying to fix that (not a Docker) expert I decided to get the refactoring done first. :)

I need to figure out why my local integration tests are broken - but that's another story. For now, I've simply configured an OkHttp provider to use with the tests - but we should definitely test this with the JDK11 provider as well.

provider() is static because that's how the builder() methods are. The original code created a new OkHttp provider on each build. I agree with you - these things should be reused, so making it static with an explicit configuration step - that's why that was part of the refactor that moved the SimpleBatch*Sender classes out of the OkHttp provider.

telemetry_core/src/main/java/com/newrelic/telemetry/metrics/SimpleMetricBatchSender.java
```diff
@@ -17,6 +17,12 @@
  */
 public class SimpleMetricBatchSender {
 
+  private static Function<Duration, HttpPoster> ctor = null;
```
jkwatson: I really don't like adding static state here. I'd much prefer, if we want this class to support something other than okhttp, that we do that as a separate step. Can you remove this part from the PR, and we'll figure out a better way to do this as a follow-on?

telemetry_core/src/main/java/com/newrelic/telemetry/spans/SimpleSpanBatchSender.java
```diff
@@ -18,6 +17,12 @@
  */
 public class SimpleSpanBatchSender {
 
+  private static Function<Duration, HttpPoster> ctor = null;
```
jkwatson: same comment as above.

integration_test/src/test/java/com/newrelic/telemetry/MetricApiIntegrationTest.java
```diff
@@ -18,11 +18,7 @@
 import com.newrelic.telemetry.exceptions.RetryWithBackoffException;
 import com.newrelic.telemetry.exceptions.RetryWithRequestedWaitException;
 import com.newrelic.telemetry.exceptions.RetryWithSplitException;
-import com.newrelic.telemetry.metrics.Count;
-import com.newrelic.telemetry.metrics.Gauge;
-import com.newrelic.telemetry.metrics.MetricBatchSender;
-import com.newrelic.telemetry.metrics.MetricBuffer;
-import com.newrelic.telemetry.metrics.Summary;
+import com.newrelic.telemetry.metrics.*;
```
jkwatson: much as it pains me to say it, I think our coding standards don't allow wildcard imports.

telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11HttpPoster.java
```diff
@@ -0,0 +1,61 @@
...
+          httpClient.send(
+              req, java.net.http.HttpResponse.BodyHandlers.ofString(Charset.defaultCharset()));
+
+      return of(response);
+    } catch (URISyntaxException | InterruptedException e) {
+      throw new IOException(e);
+    }
+  }
+
+  public static HttpResponse of(java.net.http.HttpResponse actual) {
```
jkwatson: does this need to be public?

integration_test/src/test/java/com/newrelic/telemetry/MetricApiIntegrationTest.java
jkwatson: ref here: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11HttpPoster.java
kittylyst: I was in 2 minds about that - it makes it a trifle easier to mock / test but I'm not that fussed either way - your call.

jkwatson: If we want it public, I think it needs a better name than `of`.  How about `convertToSdkResponse`?

kittylyst: I'd prefer toSdkResponse() as it conserves the loose grammatical rule that prepositions should be preferred for the names of these types of converters.

telemetry-http-okhttp/README.md
```diff
@@ -2,9 +2,6 @@ This module contains an implementation of the required http client interface for
 
 * `OkHttpPoster` : Implements the `HttpPoster` interface, using okhttp.
 
-In addition, it has two builder shims for creating telemetry senders that use okhttp: 
-`com.newrelic.telemetry.SimpleMetricBatchSender` and `com.newrelic.telemetry.SimpleSpanBatchSender`
-
```
jkwatson: this comment should probably still be here until we can revisit the `Simple*` stuff

PR #133, opened 2020-03-31
Remove lombok

kittylyst:
Lombok is deprecated by the New Relic coding standards, this PR removes the dependency from the SDK.


jkwatson:
If we're going to remove this and make everyone generate all that boilerplate, we should document in the Contributing.md how they should be doing it. Also, any class that has an equals/hashcode MUST have unit tests to verify that it is correctly implemented, otherwise bugs will result because people will forget to update the equals/hashcode when adding fields.

PR #135, opened 2020-04-02
Java11 batch senders

XiXiaPdx:
Simple batch and span sender using J11 poster.  

Is this needed?  :)

### DIFF
--- a/telemetry-http-java11/src/main/java/com/newrelic/telemetry/SimpleMetricBatchSender.java
+++ b/telemetry-http-java11/src/main/java/com/newrelic/telemetry/SimpleMetricBatchSender.java
@@ -1,0 +1,23 @@
+package com.newrelic.telemetry;
+
+import com.newrelic.telemetry.metrics.MetricBatchSender;
+import com.newrelic.telemetry.metrics.MetricBatchSenderBuilder;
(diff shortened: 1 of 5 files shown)


jkwatson:
I know this is something that @kittylyst has opinions about. What do you think of this approach, Ben? Or, would you prefer a more holistic approach to "easy mode" configuration that will auto-detect that java 11 is available?

kittylyst:
I'm in 2 minds about this - on the one hand this is mostly cut-and-paste coding that mirrors the OkHttp case, but I know that @jkwatson is not keen on the idea of static state to keep a provider in. On another, it sort of feels like this sort of Simple*Batch stuff should be part of the API. Anyone have any other ideas?

kittylyst:
Take a look at PR 136 - https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/136 for an alternative take on this.

jkwatson:
As commented in #136 , I think we should rename the java 11 ones to "Java11(Metric|Span)BatchSender", and create new versions of the OkHttp ones with the same naming convention.

jkwatson:
@XiXiaPdx I think we're going to go a slightly different route with a variant on #136 .

PR #136, opened 2020-04-04
Let's use factories

kittylyst:
Alternative approach to PR 135 - what do folks think?


telemetry_core/src/main/java/com/newrelic/telemetry/MetricBatchSenderFactory.java
```diff
@@ -4,18 +4,17 @@
...
 /**
- * A builder class for creating a MetricBatchSender that uses okhttp as the underlying http client
- * implementation.
+ * A factory interface for creating a MetricBatchSender.
  *
- * <p>Note: This class will be deprecated in the near future and be replaced with something named
- * more succinctly.
+ * <p>Concrete implementations use different HTTP providers.
  */
-public class SimpleMetricBatchSender {
```
jkwatson, line 18: If we want to replace this (and I think we do!), we should definitely start by marking this class `@Deprecated` for at least some amount of time before we delete it altogether. There are users in the wild for whom this will be a breaking change.

telemetry_core/src/main/java/com/newrelic/telemetry/SpanBatchSenderFactory.java
```diff
@@ -4,18 +4,17 @@
...
 /**
- * A builder class for creating a SpanBatchSender that uses okhttp as the underlying http client
- * implementation.
+ * A factory interface for creating a SpanBatchSender.
  *
- * <p>Note: This class will be deprecated in the near future and be replaced with something named
- * more succinctly.
+ * <p>Concrete implementations use different HTTP providers.
  */
-public class SimpleSpanBatchSender {
```
jkwatson, line 18: same comment. We can't delete this piece of the public API without at least a lot of pre-warning via a `@Deprecated` annotation and comments. I'd like to favor only accreting onto our public API, as long as the thing we're deprecating isn't actually broken/non-functional, even if it's not the API that we thing is the greatest in the long run.

telemetry_core/src/main/java/com/newrelic/telemetry/MetricBatchSenderFactory.java
kittylyst: They're not deleted - I've converted them into interfaces, so I can sidestep the static state problem.

telemetry_core/src/main/java/com/newrelic/telemetry/SpanBatchSenderFactory.java
kittylyst: They're not deleted - I've converted them into interfaces, so I can sidestep the static state problem.

jkwatson: renaming the class and replacing it with an interface is effectively deleting the class, from the user's perspective, no?

integration_test/src/test/java/com/newrelic/telemetry/MetricApiIntegrationTest.java
```diff
@@ -75,8 +75,10 @@ static void beforeClass() {
   @BeforeEach
   void setUp() throws Exception {
     mockServerClient.reset();
+    MetricBatchSenderFactory factory = d -> new OkHttpPoster(d);
```
jkwatson, line 78:
```suggestion
    MetricBatchSenderFactory factory = duration -> new OkHttpPoster(duration);
```

jkwatson:
I looked through this carefully, and I think this isn't the right approach. These factories don't provide any additional value over the existing Builder classes, which already require an API Key and a HttpPoster in order to work. 

The goal of the "Simple" classes was to provide a builder "helper" that didn't require you to specific okhttp as your http implementation if you were pulling in the okhttp sub-dependency.

I think I'd prefer to 
a) deprecate the Simple* classes
b) create a new OkHttpMetricBatchSender and Java11MetricBatchSender (and same for Spans) which look a lot like the existing Simple* classes, but then have much clearer naming.

Unfortunately, I think this factory approach will lead to confusion over the existing Builders, since they don't provide any additional functionality, and don't make it easier for users to do the simple thing simply.

telemetry_core/src/test/java/com/newrelic/telemetry/SimpleMetricBatchSenderTest.java
```diff
@@ -0,0 +1,26 @@
...
+package com.newrelic.telemetry;
+
+import static java.time.Duration.ofSeconds;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import com.newrelic.telemetry.http.HttpPoster;
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+
+class MetricBatchSenderFactoryTest {
```
jkwatson, line 14: this class name doesn't match the file name.

telemetry_core/src/test/java/com/newrelic/telemetry/SimpleSpanBatchSenderTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+  @Test
+  void testBuilders() throws Exception {
+    HttpPoster h = (url, headers, body, mediaType) -> null;
+    SpanBatchSenderFactory f = d -> h;
+
+    //    assertNotNull(f.configureWith("abc123", ofSeconds(5)));
+    //    assertNotNull(f.configureWith("abc123"));
+    //    assertNotNull(f.createBatchSender("abc123"));
+    //    assertNotNull(f.createBatchSender("abc123", Duration.ofSeconds(5)));
```
jkwatson, line 20: these look like they need to be brought back in

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
```diff
@@ -32,18 +31,31 @@
...
+            metricFactory
+                .configureWith(insightsInsertKey, Duration.of(10, ChronoUnit.SECONDS))
+                .build());
+
+    SpanBatchSenderFactory spanFactory =
+        SpanBatchSenderFactory.fromHttpImplementation(OkHttpPoster::new);
+    SpanBatchSender spanBatchSender =
+        SpanBatchSender.create(spanFactory.configureWith(insightsInsertKey).build());
+
+    // a fully customized example
```
jkwatson, line 48: I don't know if both of these need to be kept in here. Might be nice to eventually break these examples out into more specific use-cases that demonstrate the simpler usage, or the more complex options when needed.  For this PR, let's keep just the simple example and get rid of the fully-customized version.

telemetry_core/src/main/java/com/newrelic/telemetry/AbstractSenderBuilder.java
```diff
@@ -7,7 +7,7 @@
 import java.net.URI;
 import java.net.URL;
 
-public abstract class AbstractSenderBuilder<B extends AbstractSenderBuilder<B>> {
+public abstract class AbstractSenderBuilder<TB extends AbstractSenderBuilder<TB, T>, T> {
```
jkwatson, line 10: this class can go away now, correct?

Issue #72, opened 2019-09-06
put the gson classes into their original com.google.gson.stream package


jkwatson:

wpoch:
Hi, which was the intention while doing this? 
By moving the class to the original package make it conflict with the REAL class from the Gson package, so for example if I'm using Gson `2.3.1` the `jsonValue` doesn't exists, giving runtime errors, because the class you are using is in the same package of the already loaded one.

I think moving the class to the original package instead of your own, makes things more complicated.

jkwatson:
I believe the intention was to make it simpler to keep the classes up to date. I think @breedx-nr originally asked for it. However, if this is causing issues, then I don't think there's a particular strong reason not to relocate them.  What do you think @breedx-nr ?

breedx-nr:
There were two original intentions: 
1. Not have google classes moved into a `com.newrelic` package because it's misleading and somewhat misrepresentative.
2. Be able to automate license headers across the source tree (com.newrelic) without having to make onsey-twosey exceptions for non-NR classes that got brought into our namespace.
Moving them is fine (feel free to PR that), but we'll have to address license header automation if we do that.

wpoch:
Thank you both for the context, I created: https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/148

That will work for the short term, but maybe you can see if adding a shadow plugin makes sense.

PR #148, opened 2020-04-28
chore: Raname the com.google.gson package to vendored.com.google.gson to avoid conflicts with existing loaded gson versions

wpoch:
### Summary

Following the discussion opened here: https://github.com/newrelic/newrelic-telemetry-sdk-java/issues/72#issuecomment-620769410 I moved the copied classes (`JsonScope`, `JsonWriter`) from the [Gson package](https://github.com/google/gson) library to a different package than the original one `com.google.gson` to `vendored.com.google.gson`.

This will solve the issue when the `newrelic-telemetry-sdk-java` is imported in a project that is using an old version of `Gson` for example `2.3.1` which doesn't have the `JsonWriter.jsonValue` method and a runtime exception is thrown.

I have chosen the `vendored` prefix for the existing packages, I'm not proud of it, but it will work, another solution could be to configure a shadow plugin for grade to do this automatically for certain packages.

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/events/json/EventToJson.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/events/json/EventToJson.java
@@ -5,12 +5,12 @@
 
 package com.newrelic.telemetry.events.json;
 
-import com.google.gson.stream.JsonWriter;
 import com.newrelic.telemetry.events.Event;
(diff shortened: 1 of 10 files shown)


wpoch:
Here is the long term solution =) https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/149, so closing in favor of that one.

PR #149, opened 2020-04-29
feat: Use the Shadow plugin to include the Gson dependencies

wpoch:
### Summary

Configure the [Gradle Shadow Plugin](https://imperceptiblethoughts.com/shadow/) to bundle the necessary files form the `Gson` library into the final artifact.

That is achieved by including the `compile` dependencies into the `jar` file, relocating them to a different package to avoid collisions with other version of the library already in the classloader [see ](https://github.com/newrelic/newrelic-telemetry-sdk-java/issues/72#issuecomment-620769410), and then minimizing the file so only the classes being used will be present in the final artifact.

This is the real solution to the problem fixed in a really hacky way here: https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/148.


breedx-nr:
This is great, and exactly what I was thinking when I responded on #148.  I wasn't sure about shadow being able to filter/select only the couple of files we care about, though.  Do you happen to have an example of that?

telemetry_core/build.gradle.kts
```diff
@@ -21,4 +31,18 @@ dependencies {
...
\ No newline at end of file
+}
+
+tasks {
+    "shadowJar"(ShadowJar::class) {
+        classifier = ""
+        dependencies {
+            exclude(dependency("org.slf4j:slf4j-api:${Versions.slf4j}"))
+        }
+        relocate("com.google.gson", "a")
```
breedx-nr, line 42:
```suggestion
        relocate("com.google.gson", "com.newrelic.relocated")
```
Just in case something pops up and people wonder where `a` originated from.  At least this could be a hint.

breedx-nr:
Aaaaand, I see that gson is already in covered in our notices, so this is good.  I'm going to hope that the package change is ok with @wpoch and get it merged.  👍

wpoch:
Thanks @breedx-nr for merging it!
Do you know when I can expect a new released version of the library?

breedx-nr:
We don't have a date planned, and we've got some additional housekeeping work to do, but I think it's pretty close and we might expect it in early May-ish.

PR #151, opened 2020-04-29
Prepare for 050 release

kittylyst:
This adds integration tests, examples and a bit of documentation for the events functionality.


integration_test/src/test/java/com/newrelic/telemetry/EventApiIntegrationTest.java
```diff
@@ -0,0 +1,212 @@
...
+    eventBuffer.addEvent(
+        new Event("myEvent", new Attributes().put("key2", "val2"), currentTimeMillis + 1));
+    assertThrows(
+        exceptionClass,
+        () -> eventBatchSender.sendBatch(eventBuffer.createBatch()),
+        exceptionMessage);
+  }
+
+  //
+  // [{\"common\":{\"attributes\":{\"key1\":\"val1\"}},\"metrics\":[{\"name\":\"myEvent\",\"type\":\"count\",\"value\":1.0,\"timestamp\":1557180766612,\"interval.ms\":42,\"attributes\":{\"key2\":\"val2\"}}]}]"
```
jkwatson, line 203: This comment is not accurate for the Events JSON payload.

integration_test/src/test/java/com/newrelic/telemetry/EventApiIntegrationTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private static class EventPayload {
+    private final List<Map<String, Object>> metrics;
```
jkwatson, line 206: this is weird. why is there a `metrics` member in here.

jkwatson: I suspect that the payload isn't being inspected at all by mockserver, due to the `ONLY_MATCHING_FIELDS` option. We should get this payload (and the comment) whipped into shape to make sure that we're actually testing the shape of the payload, and not just the headers, etc.

kittylyst: This is exactly the same as the Span and Metrics case.

jkwatson: It's not the same. In those cases, the payload structure matches the JSON (`common` + `metrics/spans`). This payload structure is different.

jkwatson: I think when lombok got removed, the lack of getters on these little classes made it so we stopped verifying the payload structure. We need that back.

jkwatson: I put in a PR to fix this for metrics/spans.  https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/152

PR #155, opened 2020-04-30
Restore missing methods on deprecated classes.

jkwatson:
We broke backward compatibility in 0.5.0, so this should restore that.

### DIFF
--- a/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 New Relic Corporation. All rights reserved.
+ * Copyright 2020 New Relic Corporation. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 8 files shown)


telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -55,6 +55,20 @@ public TelemetryClient(
     this.eventBatchSender = eventBatchSender;
   }
 
+  /**
+   * Create a new TelemetryClient instance, with two senders. Note that if you don't intend to send
+   * one of the telemetry types, you can pass in a null value for that sender.
+   *
+   * @deprecated Use the constructor with all there senders, passing in null to the ones you don't
```
breedx-nr, line 62: s/there/three

jkwatson, line 62:
```suggestion
   * @deprecated Use the constructor with all three senders, passing in null to the ones you don't
```

jkwatson: fixed

PR #159, opened 2020-05-08
Add logging API to the SDK

jkwatson:
This addition to the SDK will be useful for non-typical logging usages, such as reporting OpenTelemetry Span Events as log entries, in context. It's not intended as an API to be used to implement high-throughput logging systems. That use-case should be reserved for log forwarder implementations.

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.newrelic.telemetry.logs;
+
+import com.newrelic.telemetry.Attributes;
+import com.newrelic.telemetry.Telemetry;
+import com.newrelic.telemetry.util.Utils;
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+public class Log implements Telemetry {
+  private final long timestamp; // in epoch ms
+  private final String message;
+  private final Attributes attributes;
+  private final String serviceName; // service.name <- goes in attributes
+  private final String logType;
+  private final String level;
+
+  private Log(LogBuilder builder) {
+    this.timestamp = builder.timestamp;
+    this.message = builder.message;
+    this.attributes = builder.attributes;
+    this.serviceName = builder.serviceName;
+    this.logType = builder.logType;
+    this.level = builder.level;
+  }
+
+  public long getTimestamp() {
+    return timestamp;
+  }
+
+  public String getMessage() {
+    return message;
+  }
+
+  public Attributes getAttributes() {
+    return attributes;
+  }
+
+  public String getServiceName() {
+    return serviceName;
+  }
+
+  public String getLogType() {
+    return logType;
+  }
+
+  public String getLevel() {
+    return level;
+  }
+
+  public static LogBuilder builder() {
+    return new LogBuilder();
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof Log)) {
+      return false;
+    }
+
+    Log log = (Log) o;
+
+    if (timestamp != log.timestamp) {
+      return false;
+    }
+    if (message != null ? !message.equals(log.message) : log.message != null) {
+      return false;
+    }
+    if (attributes != null ? !attributes.equals(log.attributes) : log.attributes != null) {
+      return false;
+    }
+    if (serviceName != null ? !serviceName.equals(log.serviceName) : log.serviceName != null) {
+      return false;
+    }
+    if (logType != null ? !logType.equals(log.logType) : log.logType != null) {
+      return false;
+    }
+    return level != null ? level.equals(log.level) : log.level == null;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = (int) (timestamp ^ (timestamp >>> 32));
+    result = 31 * result + (message != null ? message.hashCode() : 0);
+    result = 31 * result + (attributes != null ? attributes.hashCode() : 0);
+    result = 31 * result + (serviceName != null ? serviceName.hashCode() : 0);
+    result = 31 * result + (logType != null ? logType.hashCode() : 0);
+    result = 31 * result + (level != null ? level.hashCode() : 0);
+    return result;
+  }
+
+  @Override
+  public String toString() {
+    return "Log{"
+        + "timestamp="
+        + timestamp
+        + ", message='"
+        + message
+        + '\''
+        + ", attributes="
+        + attributes
+        + ", serviceName='"
+        + serviceName
+        + '\''
+        + ", logType='"
+        + logType
+        + '\''
+        + ", level='"
+        + level
+        + '\''
+        + '}';
+  }
+
+  /**
+   * A class for holding the variables associated with a Span object and creating a new Span object
+   * with those variables
+   */
+  public static class LogBuilder {
+    private long timestamp = System.currentTimeMillis();
+    private String message;
+    private Attributes attributes = new Attributes();
+    private String serviceName; // service.name <- goes in attributes
+    private String logType; // logtype <- goes in attributes
+    private String level;
+
+    public LogBuilder timestamp(long timestamp) {
+      this.timestamp = timestamp;
+      return this;
+    }
+
+    public LogBuilder level(String logLevel) {
+      this.level = logLevel;
+      return this;
+    }
+
+    public LogBuilder message(String message) {
+      this.message = message;
+      return this;
+    }
+
+    public LogBuilder attributes(Attributes attributes) {
+      this.attributes = attributes;
+      return this;
+    }
+
+    public LogBuilder serviceName(String serviceName) {
+      this.serviceName = serviceName;
+      return this;
+    }
+
+    public Log build() {
+      Utils.verifyNonBlank(message, "A message is required for a log entry");
+      return new Log(this);
+    }
+
+    @Override
+    public String toString() {
+      return "LogBuilder{"
+          + "timestamp="
+          + timestamp
+          + ", message='"
+          + message
+          + '\''
+          + ", attributes="
+          + attributes
+          + ", serviceName='"
+          + serviceName
+          + '\''
+          + '}';
+    }
+
+    public LogBuilder logType(String type) {
+      this.logType = type;
+      return this;
+    }
+
+    /** Will fill in the message with the stack trace from the Throwable. */
+    public LogBuilder stackTrace(Throwable e) {
+      Utils.verifyNonNull(e);
+      ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+      e.printStackTrace(new PrintStream(bytes));
+      this.message = bytes.toString();
+      this.logType = "stackTrace";
+      return this;
+    }
+  }
+}
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
@@ -119,7 +119,7 @@ public class Log implements Telemetry {
   }
 
   /**
-   * A class for holding the variables associated with a Span object and creating a new Span object
+   * A class for holding the variables associated with a Log object and creating a new Log object
    * with those variables
    */
   public static class LogBuilder {
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
@@ -11,6 +11,7 @@ import com.newrelic.telemetry.util.Utils;
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
 
+/** A Log instance represents a single entry in a log. */
 public class Log implements Telemetry {
   private final long timestamp; // in epoch ms
   private final String message;
@@ -28,30 +29,37 @@ public class Log implements Telemetry {
     this.level = builder.level;
   }
 
+  /** The point in time (ms since UNIX epoch) that the log entry was created. */
   public long getTimestamp() {
     return timestamp;
   }
 
+  /** The log line itself. */
   public String getMessage() {
     return message;
   }
 
+  /** Additional attributes associated with the log entry. */
   public Attributes getAttributes() {
     return attributes;
   }
 
+  /** The name of the service which produced this log entry. */
   public String getServiceName() {
     return serviceName;
   }
 
+  /** The type of log entry. This can be helpful for querying your log entries in New Relic. */
   public String getLogType() {
     return logType;
   }
 
+  /** The log level (eg. INFO, DEBUG, etc) for the log entry. */
   public String getLevel() {
     return level;
   }
 
+  /** Create a builder for building a new log entry. */
   public static LogBuilder builder() {
     return new LogBuilder();
   }
@@ -120,7 +128,7 @@ public class Log implements Telemetry {
 
   /**
    * A class for holding the variables associated with a Log object and creating a new Log object
-   * with those variables
+   * with those variables.
    */
   public static class LogBuilder {
     private long timestamp = System.currentTimeMillis();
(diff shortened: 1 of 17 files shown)


gradle.properties
```diff
@@ -4,7 +4,7 @@
 #
 
 # Here is where we manage the version
-releaseVersion=0.5.1
+releaseVersion=0.5.2-SNAPSHOT
```
jeffalder, line 7: Adding a logging API only merits a patch-level increase?

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -67,36 +69,77 @@ public static SenderConfigurationBuilder builder(String defaultUrl, String baseP
     private boolean auditLoggingEnabled = false;
     private String secondaryUserAgent;
 
+    /** Create a new Builder, with the default URL and base path. */
```
jeffalder, line 72: This javadoc (also line 60) adds no value, only further maintenance in the future.

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -67,36 +69,77 @@ public static SenderConfigurationBuilder builder(String defaultUrl, String baseP
...
+     *
+     * @return this builder;
+     */
     public SenderConfigurationBuilder apiKey(String apiKey) {
       this.apiKey = apiKey;
       return this;
     }
 
+    /**
+     * Configure the {@link HttpPoster} implementation to use.
```
jeffalder, line 89: Used for what?

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -67,36 +69,77 @@ public static SenderConfigurationBuilder builder(String defaultUrl, String baseP
...
+     *
+     * @return this builder
+     */
     public SenderConfigurationBuilder httpPoster(HttpPoster httpPoster) {
       this.httpPoster = httpPoster;
       return this;
     }
 
+    /**
+     * Configure the *full* endpoint URL for data to be sent to, including the path.
```
jeffalder, line 99:
```suggestion
     * Configure the <strong>full</string> endpoint URL for data to be sent to, including the path.
```

I think Javadoc is html, not markdown, right?

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -67,36 +69,77 @@ public static SenderConfigurationBuilder builder(String defaultUrl, String baseP
...
     }
 
+    /**
+     * Configure the endpoint URL for data to be sent to, not including the path. If the path is
+     * provided, it will be replaced with the default path.
+     *
+     * @param endpointUri A {@link URI}, path will be ignored.
+     * @return this builder.
+     */
+    public SenderConfigurationBuilder endpoint(URI endpointUri) throws MalformedURLException {
```
jeffalder, line 116: If this option is only scheme + host + port, then I would recommend that it accept three parameters. Otherwise, the choice between `endpoint` and `endpointUrl` is completely unclear. Without carefully reading the Javadoc (which should not be required), the fact that one accepts path and one does not is unexpected.

Alternative solution is to name them `endpoint(URL)` and `endpointIgnoringPath(URL)`.

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -67,36 +69,77 @@ public static SenderConfigurationBuilder builder(String defaultUrl, String baseP
...
+     *
+     * @return this builder.
+     */
     public SenderConfigurationBuilder auditLoggingEnabled(boolean auditLoggingEnabled) {
       this.auditLoggingEnabled = auditLoggingEnabled;
       return this;
     }
 
+    /**
+     * Configure a secondary User-Agent value to use when sending data.
```
jeffalder, line 133: What is a secondary user-agent? Why would I configure one?

telemetry_core/src/main/java/com/newrelic/telemetry/SpanBatchSenderFactory.java
```diff
@@ -51,24 +38,18 @@ default SpanBatchSender createBatchSender(String apiKey, Duration callTimeout) {
...
+
   /**
-   * Create a new SpanBatchSenderBuilder with your New Relic Insights Insert API key and a custom
-   * http timeout.
+   * Create an {@link SpanBatchSenderFactory} with an HTTP implementation.
    *
-   * @see <a
-   *     href="https://docs.newrelic.com/docs/apis/getting-started/intro-apis/understand-new-relic-api-keys#user-api-key">New
-   *     Relic API Keys</a>
+   * @param creator A {@link Supplier} that returns an {@link HttpPoster} implementation.
```
jeffalder, line 49: If the method's only parameter is of type `Supplier<HttpPoster>`, this javadoc reiterates the type and doesn't say where you might find one of these things.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
```diff
@@ -0,0 +1,194 @@
...
+        + logType
+        + '\''
+        + ", level='"
+        + level
+        + '\''
+        + '}';
+  }
+
+  /**
+   * A class for holding the variables associated with a Span object and creating a new Span object
```
jeffalder, line 122:
```suggestion
   * A class for holding the variables associated with a Log object and creating a new Log object
```

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
```diff
@@ -0,0 +1,194 @@
...
+   */
+  public static class LogBuilder {
+    private long timestamp = System.currentTimeMillis();
+    private String message;
+    private Attributes attributes = new Attributes();
+    private String serviceName; // service.name <- goes in attributes
+    private String logType; // logtype <- goes in attributes
+    private String level;
+
+    public LogBuilder timestamp(long timestamp) {
```
jeffalder, line 133: Here is where javadoc would be very helpful.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * A class for holding the variables associated with a Span object and creating a new Span object
+   * with those variables
+   */
+  public static class LogBuilder {
+    private long timestamp = System.currentTimeMillis();
+    private String message;
+    private Attributes attributes = new Attributes();
+    private String serviceName; // service.name <- goes in attributes
+    private String logType; // logtype <- goes in attributes
```
jeffalder, line 130: I would strike this; I don't think it adds value being a separate attribute.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/LogBatchSender.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public Response sendBatch(LogBatch batch) throws ResponseException {
+    if (batch == null || batch.size() == 0) {
+      logger.debug("Skipped sending a null or empty log batch");
+      return new Response(202, "Ignored", "Empty batch");
+    }
+    logger.debug(
+        "Sending a log batch (number of logs: {}) to the New Relic log ingest endpoint)",
+        batch.size());
+    String json = marshaller.toJson(batch);
```
jeffalder, line 60: We need to rethink the sender if we're going to implement this as-is. If the sender does _literally any logging_ you could quickly drown yourself in recursive logs without careful configuration. Think about audit logging:

1. There's an http post.
2. There's a log message with the contents of the post.
3. There's an http post with the log message with the contents of the post.
4. There's a log message with the contents of the post with the log message with the contents of the post.

gradle.properties
jkwatson: This PR includes the code from #156 which is still awaiting review. also, this is still a draft.

jkwatson:
> Even though it's nice to have the symmetry for L as we do for M, E, and T, I don't think this is the right approach. It doesn't seem useful not being integrated with any logging framework, and I think it looks at "logs as telemetry" instead of "application logs as one set of logs in an application stack".
> 
> When we implemented logs-in-context, there was a lot of debate about whether the agent should just send the log messages or we should rely on something else to do it. Matt Culmone, one of the Log PMs, was adamant that the application should _not_ be posting logs to an HTTP endpoint. There are existing solutions for this, like log forwarders, that do a far better job batching, scanning, rewinding, etc across multiple services and multiple sources than we ever could. If the user wants to consume any other type of logging, like access logs or nginx logs or non-Java logs, they would still need a log forwarder.
> 
> Before shipping this implementation, I'd encourage you to reach out to Matt or Rebecca Holzschuh to understand what they'd recommend.

Hi Jeff. My use-case for this is not acting as the primary API for logging implementations. This is to support some log-like features of OpenTelemetry. In particular, the events that can be attached to Spans are being used to log information about the processing of a given span. I'd like to experiment with writing them as log entries using the logging API.  I discussed this with Rebecca and she agreed.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/LogBatchSender.java
jkwatson: There is no expectation that this code would be used as the API for a full-fledged logging implementation, so I don't think this is a concern.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
jkwatson: I believe it does add value. It makes it very easy for people to add the logtype attribute, which is described in the logging API docs. Making the semantic conventions easy to use it a definitely value-add.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
jkwatson: Note this is still a draft PR. Before it became real, much testing and documentation would need to be added.

telemetry_core/src/main/java/com/newrelic/telemetry/SpanBatchSenderFactory.java
jkwatson: Please add this comment to #156

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
jkwatson: Please discuss in #156

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
jkwatson: Please discuss in #156

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
jkwatson: Please discuss in #156

telemetry_core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
jkwatson: please discuss in #156

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
jkwatson: javadoc added

jkwatson:
@jeffalder note, my use-case for this is supporting the "embedded" log model being discussed here:  https://github.com/open-telemetry/opentelemetry-specification/issues/622

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Will fill in the {@link #message(String)} with the stack trace from the Throwable. This will
+     * also set the {@link #logType(String)} to be "stackTrace".
+     */
+    public LogBuilder stackTrace(Throwable e) {
+      Utils.verifyNonNull(e);
+      ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+      e.printStackTrace(new PrintStream(bytes));
+      this.message = bytes.toString();
+      this.logType = "stackTrace";
```
justinfoote, line 192: Is this a convention? 

If so, it doesn't align with our other logging error semantics, here: https://github.com/newrelic/newrelic-exporter-specs/tree/master/logging#errors

jkwatson: ooh! I totally forgot there were semantics already designed for some of this stuff. Will address!

jkwatson:
> Before we go through with this, I'd like to see the exporter specs and the telemetry SDK specs updated to match.

Chicken and egg! I don't think we'd want to update the exporter specs until we've actually got an exporter using this and it's the officially sanctioned way to send span events (or whatever). Telemetry SDK, though, yes. Should we get this at least merged before we spend time writing specs for it, pre-release?

telemetry_core/src/main/java/com/newrelic/telemetry/logs/json/LogJsonCommonBlockWriter.java
```diff
@@ -0,0 +1,39 @@
...
+      return;
+    }
+    try {
+      jsonWriter.name("common");
+      jsonWriter.beginObject();
+      jsonWriter.name("attributes");
+      jsonWriter.jsonValue(attributesJson.toJson(batch.getCommonAttributes().asMap()));
+      jsonWriter.endObject();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to create span common block json", e);
```
jeffalder, line 31:
```suggestion
      throw new RuntimeException("Failed to create log common block json", e);
```

jeffalder:
@jkwatson 

> Before it became real, much testing and documentation would need to be added.
I don't see any new tests here, and it looks like this is ready to merge.

There's so much code cloned from the span side because the API shapes are so similar. Can you explain why they're cloned?

jkwatson:
> @jkwatson
> 
> > Before it became real, much testing and documentation would need to be added.
> 
> I don't see any new tests here, and it looks like this is ready to merge.
> 
> There's so much code cloned from the span side because the API shapes are so similar. Can you explain why they're cloned?

I need to add tests; just realized that there are a lot missing still. I'll be doing that ASAP. 

The code is cloned because I was moving fast to get something working. After this gets merged, I think that's definitely something to figure out how to unify.

jkwatson:
ok, added tests for json generation, and an integration test. Also got the attribute names in line with the exporter specs for logs.

jeffalder:
> The code is cloned because I was moving fast to get something working. After this gets merged, I think that's definitely something to figure out how to unify.

We already have a backlog of 16 tickets. Are you suggesting just tacking on more to the backlog? That is an issue that will never get done because it's tedious housekeeping. I feel like this is how tech debt becomes overwhelming, and we have reactions like "if we had just done this at the time ..."

I feel we have a responsibility to maintain maximally clean code. Unless there's roadmap pressure to ship this that I'm not aware of?

jkwatson:
@jeffalder do you think that this PR should also include significant refactoring of unrelated code? Do you have specific examples of where you want things refactored? I feel like refactoring should happen in a separate PR from one that adds new functionality.

telemetry_core/src/main/java/com/newrelic/telemetry/logs/Log.java
jkwatson: this has been removed, because there are no formal semantic conventions defined for it.

jkwatson:
> > Before we go through with this, I'd like to see the exporter specs and the telemetry SDK specs updated to match.
> 
> Chicken and egg! I don't think we'd want to update the exporter specs until we've actually got an exporter using this and it's the officially sanctioned way to send span events (or whatever). Telemetry SDK, though, yes. Should we get this at least merged before we spend time writing specs for it, pre-release?

@justinfoote https://github.com/newrelic/newrelic-telemetry-sdk-specs/pull/16

PR #160, opened 2020-05-12
Gradle updates and fixes to remove Gson dependency

### LINKED ISSUE #157: The SDK has a hard dependency on gson
This is a regression that happened with the 0.5.0 release. We used to include just the classes we need in the published jar; now we don't have those classes and instead depend on gson via a transitive dependency.

jeffalder:
Fixes #157 

There's really only a small necessary change to publish the shadow jar instead of the regular jar. However, because the publication is defined at the root, we have to duplicate that configuration. Extracting the methods from the root gradle was the first step.

Then I updated gradle to 6.3 from 5.4.

Then I copied the necessary code and tasks into `telemetry-core` project, and changed the line to `project.shadow.component(this)` instead of `from(components["java"])`.

Then I tweaked the shadow configuration:
1. In order to be declared as a pom-level dependency, any excluded dependencies have to end up in the `shadow` configuration. Since the `api` configuration has the same stuff, I can make `shadow` incorporate everything in `api`.
2. `compile` configuration is deprecated, so I switched it to `implementation`.
3. ~I disabled the `jar` task since it produces bad jars.~ update: I backed out this change because it was breaking the okhttp tests with NoClassDefFound ... ????
4. I changed the classifier to `archiveClassifier` because the former is deprecated

Then I made the root `build.gradle.kts` use the factored out methods from the first step.

Then I moved from `apply plugin` to `plugins`. The former is the old-style (or subproject) way of applying plugins to a build.

### DIFF
--- a/telemetry_core/build.gradle.kts
+++ b/telemetry_core/build.gradle.kts
@@ -4,10 +4,10 @@ import com.newrelic.telemetry.gradle.configuredPom
 
 plugins {
     java
+    id("java-library")
     id("com.github.johnrengelman.shadow") version "5.2.0"
 }
 
-apply(plugin = "java-library")
 apply(plugin = "maven-publish")
 apply(plugin = "signing")
 
@@ -25,9 +25,11 @@ configure<JavaPluginConvention> {
     targetCompatibility = JavaVersion.VERSION_1_8
 }
 
+configurations["shadow"].extendsFrom(configurations["api"])
+
 dependencies {
-    "api"("org.slf4j:slf4j-api:${Versions.slf4j}")
-    compile("com.google.code.gson:gson:${Versions.gson}")
+    api("org.slf4j:slf4j-api:${Versions.slf4j}")
+    implementation("com.google.code.gson:gson:${Versions.gson}")
 
     testImplementation("org.slf4j:slf4j-simple:${Versions.slf4j}")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${Versions.junit}")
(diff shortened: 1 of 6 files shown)


telemetry_core/build.gradle.kts
```diff
@@ -20,29 +25,61 @@ configure<JavaPluginConvention> {
...
         dependsOn(shadowJar)
     }
 }
+
+val useLocalSonatype = project.properties["useLocalSonatype"] == "true"
+
+configure<PublishingExtension> {
+    publications {
+        create<MavenPublication>("mavenJava") {
+            project.shadow.component(this)
```
jeffalder, line 78: This line here is one of the keys to fixing it.

telemetry_core/build.gradle.kts
```diff
@@ -20,29 +25,61 @@ configure<JavaPluginConvention> {
     targetCompatibility = JavaVersion.VERSION_1_8
 }
 
+configurations["shadow"].extendsFrom(configurations["api"])
```
jeffalder, line 28: This is the other necessary key to fix the root problem.

jeffalder:
@breedx-nr @jkwatson this is ready for review now

jkwatson:
the gradle.properties needs to be updated to publish a snapshot build. that got missed after the last release.

jkwatson:
I tested publishing this locally, pulling it down out of my local nexus and using it for doing some work, and it worked great. Thanks for cranking out this fix!!!!

Issue #170, opened 2020-05-28
Switch from azure pipelines to github actions


breedx-nr:
This is our preferred CI tooling going forward, so let's switch it over.

PR #177, opened 2020-05-29
build PRs with github actions

breedx-nr:
This should disable the PR azure action and replace it with a GitHub action.  It's not great to be split (publish still on Azure) but hopefully this is a step along the path of #170.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,22 @@
+name: Java PR build (gradle)
+
+on:
+  pull_request:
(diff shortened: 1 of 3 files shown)


breedx-nr:
I don't think there's anything else required.  That is, if you [click thru to the output](https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/177/checks?check_run_id=729134356), you'll see that they did run.

jkwatson:
@breedx-nr any reason not to merge this? Should we remove the expected check from the branch rules before?

Issue #167, opened 2020-05-26
Avoid infinite loop when getting 429 from NewRelic


vsanna:
Hi there,

I encountered one problem related with TelemetryClient, so please let me open an issue here.

I'm using micrometer-registry-new-relic(0.5.0), telemetry-core(0.4.0), telemetry(0.4.0).

## Background
When TelemetryClient is getting some kind of error from NewRelic(e.g. 429), it retries sending metrics without considering something like retryCount.
and this process is being executed on UserThread(TelmetryClient's executor), so this prevent JVM from shutting down while getting 429.

IMO, it's better for user to be able to control all of user-threads or I'd like to avoid using user-threads in libraries.

So I'd like you to consider 
- having its executor use ThreadFactory which makes its threads as daemon threads
- OR providing a constructor which we can inject our own executors from outside

## Suggestion
I hope option 2, so let me suggest option2 example..

In this line,
https://github.com/newrelic/newrelic-telemetry-sdk-java/blob/8d81d9c6a076b39b71827aa857b029156f274e1a/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java#L41

```java
private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor(r -> {
	Thread thread = new Thread(r);
	thread.setDaemon(true);
	return thread;
});
```

If you don't mind, I will open a PR for this..

## Note
I've opened another Issue in newrelic/micrometer-registry-new-relic, but the issue is covering another problem

jkwatson:
@vsanna yes, the threads definitely should be set to daemon in that executor. A PR would be welcome!

jkwatson:
BTW, it looks like our max retry count is set to 10 here:  https://github.com/newrelic/newrelic-telemetry-sdk-java/blob/master/telemetry/src/main/java/com/newrelic/telemetry/Backoff.java#L22

If that's not working, then that's definitely a bug which should be fixed ASAP.

vsanna:
@jkwatson thank you for your comments.
As I see the code, I think code works like this.

1. (when 429) RetryWithRequestedWaitException is thrown and caught in sendWithErrorHandling
- retry
- scheduleBatchSend
- sendWithErrorHandling -> start again with 1

jkwatson:
@vsanna yeah, this is definitely a bug in our 429 (RetryWithRequestedWaitException) handling in the TelemetryClient. Thanks for the report!

[...]

PR #179, opened 2020-06-03
use daemon threads in telemetryclient

vsanna:
related issue: https://github.com/newrelic/newrelic-telemetry-sdk-java/issues/167

## Background
When TelemetryClient is getting 429 from NewRelic, it is caught into loop like below:

1. sendWithErrorHandling
2. retry
3. scheduleBatchSend
4. sendWithErrorHandling

In this flow, TelemetryClient doesn't consider maxRetries.

and this flow is executed in non-daemon threads(user threads), so while getting 429, TelemetryClient prevents JVM from shutting down.

This PR fixes the latter issue.

## Change
This PR just changes executor of TelemetryClient into one which uses ThreadFactory and set its threads as daemon threads.

### DIFF
--- a/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 public class TelemetryClient {
 
   private static final Logger LOG = LoggerFactory.getLogger(TelemetryClient.class);
+  private static final long DEFAULT_SHUTDOWN_SECONDS = 3L;
 
   private final EventBatchSender eventBatchSender;
   private final MetricBatchSender metricBatchSender;
@@ -62,11 +63,12 @@ public class TelemetryClient {
       SpanBatchSender spanBatchSender,
       EventBatchSender eventBatchSender,
       LogBatchSender logBatchSender) {
-    this.metricBatchSender = metricBatchSender;
-    this.spanBatchSender = spanBatchSender;
-    this.eventBatchSender = eventBatchSender;
-    this.logBatchSender = logBatchSender;
-    this.shutdownSeconds = 3L;
+    this(
+        metricBatchSender,
+        spanBatchSender,
+        eventBatchSender,
+        logBatchSender,
+        DEFAULT_SHUTDOWN_SECONDS);
   }
 
   /**
@@ -223,7 +225,8 @@ public class TelemetryClient {
         executor.shutdownNow();
       }
     } catch (InterruptedException e) {
-        LOG.error("interrupted graceful shutdown", e);
+      Thread.currentThread().interrupt();
+      LOG.error("interrupted graceful shutdown", e);
     }
   }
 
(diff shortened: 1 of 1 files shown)


vsanna:
@breedx-nr sorry for having you waiting for a long time..
please review on you side, this is final version of this PR from our side.

since you gave us reviews, I changed a little to enable developers to choose threads as user-threads or daemon-threads.
I believe that generally no one wants to use user-threads in library, but I'm concerned that someone will encounter other issue because its default behavior changes.

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -36,14 +36,16 @@
 public class TelemetryClient {
 
   private static final Logger LOG = LoggerFactory.getLogger(TelemetryClient.class);
+  private static final long DEFAULT_SHUTDOWN_SECONDS = 3L;
```
breedx-nr, line 39: Since this is seconds, let's make this an integer instead of a long.
(Sorry I didn't catch this the first time around)

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -58,10 +60,32 @@ public TelemetryClient(
       SpanBatchSender spanBatchSender,
       EventBatchSender eventBatchSender,
       LogBatchSender logBatchSender) {
+    this(metricBatchSender, spanBatchSender, eventBatchSender, logBatchSender, DEFAULT_SHUTDOWN_SECONDS, DEFAULT_IS_DAEMON);
+  }
+
+  /**
+   * Create a new TelemetryClient instance, with three senders and seconds to wait for shutdown.
```
breedx-nr, line 67:
```suggestion
   * Create a new TelemetryClient instance, with four senders and seconds to wait for shutdown.
```

breedx-nr:
Oh, this repo also enforced google java code formatting.  Please run `./gradlew googleJavaFormat` and commit (I wish this was automated).

vsanna:
@breedx-nr sorry, now run it and pushed! 🙇

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
vsanna: sorry, fixed

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
vsanna: fixed this 🙇 
the reason I used long is awaitTermination's signature seems to require long for the length of any time unit. `awaitTermination(long timeout, TimeUnit unit)`

https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/ExecutorService.html#awaitTermination-long-java.util.concurrent.TimeUnit-

vsanna:
@breedx-nr sorry for having you waiting for a long time..
this is the first PR form our company to OSS, so now getting approval form our internal legal team.
please give me more 3 business day 🙇

breedx-nr:
@vsanna no rush, totally understand.  While we're waiting, can you rebase for a clean merge?

vsanna:
@breedx-nr I've rebased and agreed with your cla, please check your side as well 🙇

breedx-nr:
Looks good.  Thanks again for the contribution!

PR #180, opened 2020-06-03
Add easy support for error message and class

### LINKED ISSUE #165: Provide a helper method for adding error.message
Currently, SpanBuilder.withError() marks that a span is erroneous but does not take an error message as argument. The error message has to be separately added as an Attribute.

breedx-nr:
This resolves #165

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
@@ -253,8 +253,7 @@ public final class Span implements Telemetry {
     }
 
     /**
-     * Call this to indicate that the span contains an error condition
-     * with the given message.
+     * Call this to indicate that the span contains an error condition with the given message.
      *
      * @param errorMessage The error message to be placed into the "error.message" attribute
      * @return The SpanBuilder instance with the error field set to true
@@ -264,8 +263,8 @@ public final class Span implements Telemetry {
     }
 
     /**
-     * Call this to indicate that the span contains an error condition
-     * with the given message and class.
+     * Call this to indicate that the span contains an error condition with the given message and
+     * class.
      *
      * @param errorMessage The error message to be placed into the "error.message" attribute
      * @param errorClass The error class to be placed into the "error.class" attribute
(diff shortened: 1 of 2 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/spans/Span.java
```diff
@@ -249,7 +249,35 @@ public SpanBuilder parentId(String parentId) {
...
+     * class.
+     *
+     * @param errorMessage The error message to be placed into the "error.message" attribute
+     * @param errorClass The error class to be placed into the "error.class" attribute
+     * @return The SpanBuilder instance with the error field set to true
+     */
+    public SpanBuilder withError(String errorMessage, String errorClass) {
       this.error = true;
+      if (errorMessage != null) {
+        attributes.put("error.message", errorMessage);
```
jkwatson, line 276: this is the first time we've had the builder manipulate the attributes. If we want to go down this route, we should probably 
a) deprecate the `attributes` bulk-adding method and
b) add a method to add an attribute directly.

In any case, we need to document in both of these new methods that if you call `attributes` after calling these, that you will lose what you just set.

breedx-nr: Agreed that it's the first time that this builder adds attributes, but why do you think we should deprecate the bulk add?  The single add seems useful, but we haven't had an immediate need yet?  

I can definitely add a blurb about overwriting.

jkwatson: I was just thinking that having 2 conflicting things was a bit confusing. I'm ok with it, if you think documentation is good enough.

jkwatson: What do our exporters do today? Would they be simpler if they were able to add the attributes directly to the builder, rather than having to build the attributes, then set them?

breedx-nr: the OT one builds them together and passes them in https://github.com/newrelic/opentelemetry-exporter-java/blob/f65f3451261ba2bfcf66909b4bc9f524ac06537e/opentelemetry-exporters-newrelic/src/main/java/com/newrelic/telemetry/opentelemetry/export/SpanBatchAdapter.java#L64

jkwatson: But would it be simpler just to use the SpanBuilder for that purpose? I don't know, honestly...just thought it would be a good question to ask.

breedx-nr: I'm ok with this if you are.  It seems in line with what was originally suggested and I don't find it offensive.  :)

PR #191, opened 2020-06-18
Fix the default metric URL to point at the metric API endpoint

jkwatson:
And, release 0.6.1


telemetry_core/src/main/java/com/newrelic/telemetry/metrics/MetricBatchSender.java
```diff
@@ -23,7 +23,7 @@
 public class MetricBatchSender {
 
   private static final String METRICS_PATH = "/metric/v1";
-  private static final String DEFAULT_URL = "https://trace-api.newrelic.com/";
+  private static final String DEFAULT_URL = "https://metric-api.newrelic.com/";
```
breedx-nr, line 26: 👍

Issue #176, opened 2020-05-29
Have the TelemetryClient be less noisy after being shut down


jkwatson:
Right now, if you shut down the TelemetryClient, and some other background task is trying to send data with it, you will get a big `RejectedExecutionException`. I think that the client should check to see if it has been shut down and just log a message (INFO?) that it is not going to send the data.

breedx-nr:
There's always going to be a race with this, though, right?  Like, it could get shut down at the same time that the check happened?  Maybe one `RejectedExecutionException` is better than the current behavior?

jkwatson:
I was just thinking you could ask the Executor if it was up & running before doing the submit. But, yes, there will still be a small race condition that I think we can definitely live with.

tbradellis:
pr #193

jkwatson:
I don't think this solved the problem, because we'll still get the noisy stack trace from the logger.

tbradellis:
Do we want to just not show the stack trace and log that there was an issue? And/or, do a check to see if the executor isTerminated in `scheduleBatchSend` just before attempting to schedule?

PR #193, opened 2020-06-20
Have the TelemetryClient be less noisy after being shut down 

tbradellis:
This is a start for issue #176 

I think this small change is all that is needed. 
If the TelemetryClient executor is shutdown 
log that the batch was not sent.
log some info about the batch
return before attempting to send.

I will also add a test to this pr.

### DIFF
--- a/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
@@ -17,15 +17,14 @@ import com.newrelic.telemetry.metrics.MetricBatch;
 import com.newrelic.telemetry.metrics.MetricBatchSender;
 import com.newrelic.telemetry.spans.SpanBatch;
 import com.newrelic.telemetry.spans.SpanBatchSender;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This class should be the go-to spot for sending telemetry to New Relic. It includes the canonical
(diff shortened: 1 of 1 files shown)


telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -169,6 +169,10 @@ private void scheduleBatchSend(
 
   private void sendWithErrorHandling(
       BatchSender batchSender, TelemetryBatch<? extends Telemetry> batch, Backoff backoff) {
+    if (executor.isShutdown()) {
+      LOG.info("TelemetryClient background Executor is shutdown. Batch not sent", batch.toString());
```
jkwatson, line 173: I think if we just throw the toString into the log message, it'll end up being potentially huge log line. How about just logging the number of items in the batch, and the type of batch it is (`batch.getClass().getSimpleName()`)?

jkwatson:
Looks good as a start (just one litttle comment), and needs a rebase.

tbradellis:
After looking at it more, I couldn't think of a way to pull off a test for it.
Switched to use isTerminated and added a try/catch when the batch is scheduled.   
Changed log entry to get just simple class name.

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -164,11 +165,21 @@ private void scheduleBatchSend(
...
+    try {
+      executor.schedule(() -> sendWithErrorHandling(sender, batch, backoff), waitTime, timeUnit);
+    } catch (RejectedExecutionException e) {
+      LOG.error("Problem scheduling batch.", e);
+    }
   }
 
   private void sendWithErrorHandling(
       BatchSender batchSender, TelemetryBatch<? extends Telemetry> batch, Backoff backoff) {
+    if (executor.isTerminated()) {
```
breedx-nr, line 177: I don't think we want to check the executor here.  If we have reached this line, it means that the executor has invoked us, and we should be ok to send the batch.  If the timing is perfectly wrong, I think this could cause us to lose data.

breedx-nr:
Thanks!  I think this is a good change.

PR #197, opened 2020-07-07
When an IOException occurs, have the BatchDataSender only log the message

jkwatson:
But, include the exception in the one that is rethrown in case someone higher up wants to log the details.

This should quiet things down a bit when there are transient network blips.

### DIFF
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 New Relic Corporation. All rights reserved.
+ * Copyright 2020 New Relic Corporation. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 package com.newrelic.telemetry.transport;
@@ -149,8 +149,9 @@ public class BatchDataSender {
       }
     } catch (IOException e) {
       logger.warn(
-          "IOException while trying to send data to New Relic. Batch retry recommended.", e);
-      throw new RetryWithBackoffException();
+          "IOException (message: {}) while trying to send data to New Relic. Batch retry recommended.",
+          e.getMessage());
+      throw new RetryWithBackoffException(e);
     }
   }
 
--- a/telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
+++ b/telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
@@ -151,7 +151,11 @@ public class BatchDataSender {
       logger.warn(
           "IOException (message: {}) while trying to send data to New Relic. Batch retry recommended.",
           e.getMessage());
-      throw new RetryWithBackoffException(e);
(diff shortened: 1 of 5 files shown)


telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
```diff
@@ -149,8 +149,9 @@ private Response sendPayload(byte[] payload)
       }
     } catch (IOException e) {
       logger.warn(
-          "IOException while trying to send data to New Relic. Batch retry recommended.", e);
-      throw new RetryWithBackoffException();
+          "IOException (message: {}) while trying to send data to New Relic. Batch retry recommended.",
+          e.getMessage());
+      throw new RetryWithBackoffException(e);
```
breedx-nr, line 154: Maybe nitpicky/subtle -- but the messaging inside the exception now suggests that the API recommended exponential backoff, which isn't really necessarily true in the generic IOException, right?

telemetry_core/src/test/java/com/newrelic/telemetry/transport/BatchDataSenderTest.java
```diff
@@ -65,4 +69,31 @@ void testSecondaryUserAgent() throws Exception {
 
     assertEquals(new Response(202, "OK", "yepyep"), response);
   }
+
+  @Test
+  void testCapturingCaseOfIoException() throws Exception {
+    URL endpointURl = new URL("http://foo.com");
```
breedx-nr, line 75: Let's try and standardize on `example.com` for urls.

telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
jkwatson: Fair, although that hasn't actually changed in this case...it was doing it all along. :)

telemetry_core/src/test/java/com/newrelic/telemetry/transport/BatchDataSenderTest.java
jkwatson: I'll update all the tests in this file for that.

telemetry_core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
jkwatson: Should the exception case have a different message, do you think?

breedx-nr: It probably does, something similar to what's being logged...just to avoid the incorrect suggestion about the API.

jkwatson: ok, check it out now

telemetry_core/src/test/java/com/newrelic/telemetry/transport/BatchDataSenderTest.java
jkwatson: done

breedx-nr:
Thanks thanks

PR #202, opened 2020-07-14
generate a version file and read it at runtime.

### LINKED ISSUE #144: Make the version-detection logic more robust
From looking at user-agent metadata, it looks as though the SDK is being pulled into a fat jar, and we're reading the manifest of the fat jar, rather than the original SDK jar.  

We're ending up with user-agent strings that look like: "NewRelic-Java-TelemetrySDK/20.6.7029", which is definitely not a version that we've released.

This issue will also apply to all of our java-based exporters that are reading versions from the jar manifest.

breedx-nr:
This should resolve #144

### DIFF
--- a/telemetry_core/build.gradle.kts
+++ b/telemetry_core/build.gradle.kts
@@ -63,9 +63,24 @@ tasks {
         relocate("com.google.gson", "com.newrelic.relocated")
         minimize()
     }
+    val propertiesDir = "build/generated/properties"
+    val versionFilename = "telemetry.sdk.version.properties"
+    sourceSets.get("main").output.dir(mapOf("builtBy" to "generateVersionResource"), propertiesDir)
+    register("generateVersionResource") {
+        outputs.file(File("$propertiesDir/$versionFilename"))
+        doLast {
+            val folder = file(propertiesDir)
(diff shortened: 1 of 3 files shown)


telemetry_core/build.gradle.kts
```diff
@@ -63,9 +63,24 @@ tasks {
...
+            val folder = file(propertiesDir)
+            folder.mkdirs()
+            val propertiesFile = File(folder.getAbsolutePath(), versionFilename)
+            propertiesFile.writeText("telemetry.sdk.version=${project.version}")
+        }
+    }
+
     build {
         dependsOn(shadowJar)
+        dependsOn("generateVersionResource")
```
jkwatson, line 81: does the order matter here? Just want to make sure that the shadow jar has the version file in it.

breedx-nr: It might.  Lemme fiddle with it and fix the formatting.

breedx-nr: I tried several different build sequences and couldn't get it to NOT generate and include the properties file, so I think it's ok this way.

breedx-nr:
ready for another look

PR #205, opened 2020-07-16
use new gradle syntax

### LINKED ISSUE #199: README uses old style gradle compile()
We are using newer versions of gradle and should switch to the `implementation()` syntax in the docs (README)

breedx-nr:
This resolves #199

I hope that `api()` is correct for telemetry while `implementation()` is correct for the okhttp lib support.  🤷 

Cribbed it from the OT exporter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Maven dependencies:
 Gradle dependencies: 
 
 ```
-api("com.newrelic.telemetry:telemetry:0.6.1")
+implementation("com.newrelic.telemetry:telemetry:0.6.1")
 implementation("com.newrelic.telemetry:telemetry-http-okhttp:0.6.1")
 ```
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Maven dependencies:
 Gradle dependencies: 
 
 ```
-compile("com.newrelic.telemetry:telemetry:0.6.1")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,8 +43,8 @@ Maven dependencies:
 Gradle dependencies: 
 
 ```
-compile("com.newrelic.telemetry:telemetry:0.6.1")
-compile("com.newrelic.telemetry:telemetry-http-okhttp:0.6.1")
+api("com.newrelic.telemetry:telemetry:0.6.1")
+implementation("com.newrelic.telemetry:telemetry-http-okhttp:0.6.1")
```
jkwatson, line 47: this change feels weird to me, since we don't know if the user of this library will expose our APIs in theirs or not. But, I also know that `compile` is deprecated, so I honestly have no idea what we should put here!  :)

breedx-nr: Maybe @jeffalder has an opinion on this?

jeffalder: I would, by default, recommend `implementation`. If they want to expose our classes from one project to another, the compiler will complain, and then they can switch their own dependency to `api`. When I managed the change for the agent, I changed _only_ the dependencies that required exposure. There were only 11 `api` usages versus 1045 instances of `implementation`.

breedx-nr: I'm happy with that.

PR #206, opened 2020-07-18
add simple factories in the http implementations.

### LINKED ISSUE #147: Make it easier to choose an http implementation and get a TelemetryClient
Possibilities:

Have the okhttp & java11 modules depend on the `telemetry` module, with a factory class that builds the telemetry client in one method call.

breedx-nr:
This resolves #147 

I'm not sure how much I like the names of the factory classes and am open to better ideas.

### DIFF
--- a/telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11Http.java
+++ b/telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11Http.java
@@ -1,12 +1,7 @@
 package com.newrelic.telemetry;
 
-import com.newrelic.telemetry.http.HttpPoster;
-import java.util.function.Supplier;
-
 /** Simple way to create a TelemetryClient that is backed by a Java11+ based HTTP poster. */
 public class Java11Http {
-
-  private static final Supplier<HttpPoster> supplier = Java11HttpPoster::new;
 
   /**
    * Create a new TelemetryClient with the given apiKey and default configuration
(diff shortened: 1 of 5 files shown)


telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11Http.java
```diff
@@ -0,0 +1,29 @@
+package com.newrelic.telemetry;
+
+import com.newrelic.telemetry.http.HttpPoster;
+import java.util.function.Supplier;
+
+/** Simple way to create a TelemetryClient that is backed by a Java11+ based HTTP poster. */
+public class Java11Http {
+
+  private static final Supplier<HttpPoster> supplier = Java11HttpPoster::new;
```
jkwatson, line 9: It feels a little odd to have this up here in a final var. Fewer characters in the file just to use it at the call site, I think.

telemetry-http-okhttp/src/main/java/com/newrelic/telemetry/OkHttp.java
```diff
@@ -0,0 +1,29 @@
+package com.newrelic.telemetry;
+
+import com.newrelic.telemetry.http.HttpPoster;
+import java.util.function.Supplier;
+
+/** Simple way to create a TelemetryClient that is backed by an OkHttp based HTTP posters. */
+public class OkHttp {
+
+  private static final Supplier<HttpPoster> supplier = OkHttpPoster::new;
```
jkwatson, line 9: ditto

telemetry-http-java11/src/main/java/com/newrelic/telemetry/Java11Http.java
breedx-nr: ok, that's some reasonable duplication.

PR #208, opened 2020-07-22
Update split for events

### LINKED ISSUE #150: Revisit the retry-with-splitting logic
Post 0.5.0 we should revisit how the 413 response code (retry with split) is handled. It should probably be punted up to TelemetryClient

breedx-nr:
This resolves #150 

Remove the handling of the specific retry inside the EventBatchSender and let it bubble up to the TelemetryClient where it can be retried consistently like the other telemetry types.


breedx-nr:
I think so yeah, it's true for the other types as well.  I think the timestamps inside each type prevent the need for in-order sending.....but I can't guarantee it in the backend.

PR #209, opened 2020-07-22
Upgrading okHTTP to 4.8.0

mjarvie:
Upgrading okHTTP to the latest (4.8.0) to work around an issue with cell load shifting in Infinite Tracing. [The upgrade guide](https://square.github.io/okhttp/upgrading_to_okhttp_4/) mentions that while it's now written in Kotlin, the API is stable fro 3.x -> 4.x.

### DIFF
--- a/telemetry-http-okhttp/build.gradle.kts
+++ b/telemetry-http-okhttp/build.gradle.kts
@@ -1,5 +1,5 @@
 private object Versions {
-    const val okhttp = "3.14.1"
+    const val okhttp = "4.8.0"
     const val junit = "5.3.1"
     const val jsonassert = "1.5.0"
 }


jeffalder:
@mjarvie How does this work around the error, exactly? This is using the HTTP trace endpoint, not the gRPC endpoint.

mjarvie:
I'm testing what happens to a toy app using the OpenTelemetry exporter that gets its load shifted to a new cell. During a shift, the DNS entry describing the cell is updated to point to a new one; even when the connection between the app and the cell is severed, in this version of OkHttp the updated DNS entry is never read - it only shifts to the new cell on JVM restart. I debugged this for awhile but couldn't find any configuration options that fixed this, but luckily an update fixes it.

My ultimate goal here is to cut a new release of the SDK and include that in the OT exporter.

breedx-nr:
It looks to be API compatible and since 4.x is the current development line I don't see any reason why we wouldn't want to stay current.

PR #214, opened 2020-08-04
Build telemetry-all from sourceSets rather than project dependencies

ysb33r:
Instead of trying to manipulate the various subproject JARs into
a single JAR that could be a module, this approach collects the
source directories from those subprojects and creates a special
source set within the project. in reality everything gets built again,
but now everything should be within a single toplevel package.

### DIFF
--- a/telemetry-all/build.gradle.kts
+++ b/telemetry-all/build.gradle.kts
@@ -1,4 +1,6 @@
 private object Versions {
+        const val slf4j = "1.7.30"
+        const val gson = "2.8.6"
 }
 
 plugins {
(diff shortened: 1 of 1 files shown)


jkwatson:
@ysb33r This is not against our `main` branch and updates a file that doesn't exist on `main`.

ysb33r:
@jkwatson | know about that. I was working with @

> @ysb33r This is not against our `main` branch and updates a file that doesn't exist on `main`.

Hi @jkwatson I am aware of that. I'm working with @kittylyst to restructure that part of the build.

jkwatson:
great! sorry for the noise!

PR #218, opened 2020-08-18
Add x-request-id header

### LINKED ISSUE #200: Add x-request-id header to outbound requests.
A new header named `x-request-id` with a uuid4 value should be sent on every request

| Header Name | Header Value | Code Example |
| ----------- | ------------ | ------------ |
| x-request-id | A [version 4 UUID string](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) | [str(uuid.uuid4())](https://docs.python.org/3/library/uuid.html#uuid.uuid4) |

See https://github.com/newrelic/newrelic-telemetry-sdk-specs/pull/17 for details.

breedx-nr:
This resolves #200 

Note: In order to appear more consistent with the other headers, I left the casing mixed.  I assume that this is unimportant and that most frameworks will lowercase it anyway.  It just seemed too inconsistent to leave it lowercase here.

### DIFF
--- a/integration_test/src/test/java/com/newrelic/telemetry/EventApiIntegrationTest.java
+++ b/integration_test/src/test/java/com/newrelic/telemetry/EventApiIntegrationTest.java
@@ -89,8 +89,8 @@ class EventApiIntegrationTest {
   }
 
   @Test
-  @DisplayName("Low Level SDK can send a metric batch and returns a successful response")
-  void testSuccessfulMetricSend() throws Exception {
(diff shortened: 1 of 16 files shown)


jkwatson:
can we add a bit to the integration tests to verify this behavior as well?

breedx-nr:
Yeah probably!  I forgot we had those.  :). Let me have a look.

breedx-nr:
how's that look?

PR #223, opened 2020-08-24
Provide some help for modular applications

kittylyst:
* Produce a single aggregated artifact in addition to the split ones
* Declare an automatic module name


telemetry-all/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+                "../telemetry-http-java11/src/main/resources"
+        ))
+    }
+}
+dependencies {
+    api("org.slf4j:slf4j-api:${Versions.slf4j}")
+    implementation("com.google.code.gson:gson:${Versions.gson}")
+}
+
+//plugins.withType<JavaPlugin>().configureEach {
```
jkwatson, line 36: let's not put in commented-out code, if we can help it. If this is important to leave here, it needs a comment to describe why, and what it would take to get it uncommented (and an issue created to track that work).

telemetry-all/README.md
```diff
@@ -0,0 +1,6 @@
+### Telemetry Combined Module
+
+This module provides high-level APIs for sending telemetry to New Relic.
```
jkwatson, line 3: It's probably worth mentioning what this includes, and what it doesn't (for example, that it includes a java 11 http client sender)

breedx-nr:
Oops, this was approved but never merged..and this seems pretty harmless to me, even if I'm not fully clear on the intent.  Going to merge.

PR #227, opened 2020-09-20
Fix modular build issues for telemetry-all

ysb33r:
* Renamed directory `telemetry_core` -> `telemetry-core`
* Cleaned up subprojects fox consistency
* Keep all versions in gradle.properties so that they are consitent across the project
* Make `telemetry-all` a fully modular build.

// cc @kittylyst


kittylyst (approved):
LGTM. 🚢

telemetry-all/build.gradle.kts
```diff
@@ -1,44 +1,44 @@
-private object Versions {
-        const val slf4j = "1.7.30"
-        const val gson = "2.8.6"
-}
-
 plugins {
     java
     id("java-library")
+    id( "org.ysb33r.java.modulehelper")
```
breedx-nr, line 4: I am hesitant to have our production build path requiring use of a nonstandard, external 3rd party plugin (even if @ysb33r did create it).  Perhaps we can just include the plugin in this project instead?

build.gradle.kts
```diff
@@ -7,6 +7,8 @@ plugins {
     id("maven-publish")
     id("signing")
     id("com.github.sherter.google-java-format") version "0.8"
+    id( "org.ysb33r.java.modulehelper") version("0.9.0") apply false
+    id("com.github.johnrengelman.shadow") version ("5.2.0") apply false
```
breedx-nr, line 11: I don't understand why these would be included here but not applied and then shadow isn't even applied in subprojects.  If the shading is being removed, shouldn't the use of this plugin also be removed?

breedx-nr (changes requested):
I am not sure that we can remove the gson shading, because of #149 and #72 and [some related fallout](https://github.com/newrelic/opentelemetry-exporter-java/issues/42).  This was user-submitted, and solves a very specific case where the user needs a different version of gson than `telemetry-core` needs.  As such, we decided to shade it into the jar.

telemetry-all/build.gradle.kts
ysb33r: What do you define as standard vs non-standard?

build.gradle.kts
ysb33r: Ah right, if we forego the shading I'll have to remove the shadow plugin - I missed that.

> I don't understand why these would be included here but not applied

The reason is of maintenance. Declare all of the Gradle plugins with their versions in one place which is the root project. Then only apply them selectively in the subproject that requie them. This approach suffuce for probably 98%+ os use cases. It is rare to reuire two different version of a plugin in the same multiproject.

ysb33r:
@breedx-nr If we need to keep the gson shading, we'll need to change the build as to build a modular telemetry-all without the shading and build telemetry-core with shading. it will make the build more complex of course, but it is achievable.

kittylyst:
> I am not sure that we can remove the gson shading, because of #149 and #72 and [some related fallout](https://github.com/newrelic/opentelemetry-exporter-java/issues/42). This was user-submitted, and solves a very specific case where the user needs a different version of gson than `telemetry-core` needs. As such, we decided to shade it into the jar.

OK. So instead we need to ship a non-modular set of split jars *with* shading and a single, modular jar *without* shading. The latter can possibly break any customers because no-one's using it yet!

telemetry-all/build.gradle.kts
kittylyst: To follow up on this - we've agreed that the ModuleHelper plugin will migrate into the New Relic org in the short-medium term. So I think it's fine to depend upon it in its current location for now - but will clean it up once the migration has happened.

ysb33r:
I'm closing this and creating a new PR

PR #228, opened 2020-09-28
Fix the default event endpoint to be the correct one.

### LINKED ISSUE #224: Default ingest endpoint for events is incorrect
Check this out:
https://github.com/newrelic/newrelic-telemetry-sdk-java/blob/main/telemetry_core/src/main/java/com/newrelic/telemetry/events/EventBatchSender.java#L25

The events are, by default, being sent to the trace api, which is incorrect.

breedx-nr:
This fixes #224 

Change the default event endpoint to be the right one, which is https://insights-collector.newrelic.com/v1/accounts/events


jasonjkeller (approved):
👍

PR #246, opened 2020-10-21
update changelog for 0.9.0 release

XiXiaPdx:


XiXiaPdx:
@breedx-nr  I think I'm going to update the release process info a bit. 

To release 0.9.0, I need to do that through Github releases. This will trigger the publish action, and stage the artifact in Sonatype for manual release.  Please let me know if I am misunderstanding the process here

CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Coming soon
-- remove several deprecated APIs
+
+## [0.9.0] - 2020-10-23
+- Remove several deprecated APIs.
```
breedx-nr, line 10: I don't think we actually removed these yet.  Let's keep it in the "coming soon".

breedx-nr (changes requested):
Version 0.9.0 is correct.  After this is released we will bump to 0.10.0.
Thanks.

breedx-nr (approved):

Issue #189, opened 2020-06-17
Don't retry on HTTP 429 forever


breedx-nr:
This is a follow-up from #167 .

The fix for #167 made threads daemon, but the root problem still remains:  429 errors returned from the endpoints cause the TelemetryClient to retry forever.  

There should be an upper limit to the number of consecutive retries.  I'm not sure what makes a sane default, because it might be dependent on the wait-after intervals commonly returned on 429.

breedx-nr:
Because the 429 may be accompanied by a backoff time, we think that trying forever is probably the right approach.  The previous change to daemon threads should also allow the JVM to be shut down cleanly, even when the 429 loop is in progress.

Feel free to reopen if you would like to revisit this.

breedx-nr:
We should come back to this, because we will continue to buffer new telemetry during periods of 429.  At some point we should probably start dropping data.

XiXiaPdx:
429 are accompanied with a backoff time and there is also a limiting scheduler added in #251 . This issue is closed and please reopen if there is a need.

PR #251, opened 2020-11-04
Add limiting scheduler

### LINKED ISSUE #249: Memory leak in OT Tracing 0.4.1 / 0.6.1
We use Open Telemetry tracing in our product and log spans on almost all request to our Community web application. One of our production sites (app name: xxxxxxx.prod, app id: xxxxxxxxx) on one of our hosts (hostname.domain.com) starting seeing a large spike in memory use on 10/15 until we turned of the open telemetry tracing just before 10/21. We took a memory snapshot and found about 10.77 GB was being used by the class com.newrelic.telemetry.TelemetryClient.

It seems to have backed up 10.77 GB of RunnableScheduleFutures with com.newrelic.telemetry.spans.SpanBatch objects to send to New Relic in a ScheduleThreadPoolExecutore.DelayedWorkQueue.

We checked out logs and found steadily increasing log messages from 10/14/2020 23:57:52.947 -0700 to 10/20/2020 23:55:49.481 -0700 (when we turned off open telemetry tracing for this community web app) that looked like this:

2020-10-15 08:57:52,947 +0200 [8-thread-1] INFO [cid=, tx=, rh=, userId=] newrelic.telemetry.TelemetryClient - Metric batch sending failed. Backing off 0 MILLISECONDS

We currently have open telemetry tracing turned off for this app, but we'd like to turn it back on and make sure that it doesn't happen again (or on any of our other community apps). Can you see what might have happened to cause this issue?

I can provide any further info that's needed. thanks!

breedx-nr:
Fixes #249.  Relates to (and probably mitigates) #189.

Would definitely like a critical eye on this one please 🙏 .  

In the event of a network outage or other long-standing ingest failure, we need a way to prevent the SDK from forever accumulating data (and eventually blowing the process memory limit).  Limiting the number of retries (like mentioned in #189) is one idea, but doesn't necessarily solve the problem if data continues to be scheduled for delivery at a rate that's faster than the retry+drop rate.  This is challenging to calculate (and also challenging to reason about!).

I thought it would be nice to keep track of how much data is "in-flight" at a given time and to put some large-but-reasonable overall limiter on that.  The scheduled executor service doesn't provide a way to see the number of tasks currently scheduled, so I came up with something else.  We all want a byte count...which is often unrealistic, so hopefully this pragmatic step solves the need.

This approach has the `TelemetryClient` use a new class called `LimitingScheduler` which delegates to a `ScheduledExecutorService` and leverages a counting `Semaphore` to limit the telemetry scheduled for delivery.  The size of a `TelemetryBatch` corresponds to the number of permits asked of the `Semaphore`.  This max limit applies universally across all MELT (metrics, events, logs, traces) types.  When a given `TelemetryBatch` has completed, the permits are released from the `Semaphore`.  👉  Because there's some edge casey error conditions in there, I would love some extra attention to help ensure that permits cannot be lost/leaked.

One caveat: When choosing a max size for the buffer, one must consider the retry-with-split condition.  In the version submitted here, the reschedule of split batches happens within the context of the scheduled task, and so the failed batch has not yet been "removed" from the tracking scheduler.  Because of that, the max size of the limit should be at **least** `2 * L` where L is the largest expected batch size.  Bonus caveat: If the subbatches also fail with retry, we have an even more complicated scenario that has not been considered here (tho I think that the single threaded nature of our scheduler handles this gracefully...it probably requires more max headroom).  

I picked 1,000,000 telemetry points as a default, which can be overridden by the user.  Please chime in if that seems too high or low.  I think it might be an order of magnitude too high, but


mikelaspina:
This is probably beyond the scope of this PR, but as a service owner (and given the time value of telemetry data) I generally prefer to drop old data vs new data given the choice.

kittylyst (approved):
LGTM.

kittylyst:
> This is probably beyond the scope of this PR, but as a service owner (and given the time value of telemetry data) I generally prefer to drop old data vs new data given the choice.

I agree - but that probably is too much to get done in this PR. Let's merge this and open an issue for the "drop old data" behaviour.

breedx-nr:
> This is probably beyond the scope of this PR, but as a service owner (and given the time value of telemetry data) I generally prefer to drop old data vs new data given the choice.

@mikelaspina Thanks for that perspective...I appreciate that.  I think there's a trade off: The old data could potentially hold the keys to _why_ you started having problems sending your data.  If and when things recover, it might be nice to see what happened right before the outage/problem.

I do think it's outside the scope of this PR, but I can open an issue to address this at a future time.

jasonjkeller (approved):
LGTM

PR #253, opened 2020-11-07
shutdown  executor

XiXiaPdx:
shutdown the testClass scheduler to make sure the scheduled command _could_ run (triggering assertions) if for some reason we don't respect the limit.

### DIFF
--- a/telemetry/src/test/java/com/newrelic/telemetry/LimitingSchedulerTest.java
+++ b/telemetry/src/test/java/com/newrelic/telemetry/LimitingSchedulerTest.java
@@ -15,6 +15,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


telemetry/src/test/java/com/newrelic/telemetry/LimitingSchedulerTest.java
```diff
@@ -55,10 +56,7 @@ void testScheduleFailsWhenOverCapacity() throws Exception {
     boolean result2 =
         testClass.schedule(
             6,
-            () -> {
-              seen.put("2", "");
-              latch.countDown();
-            },
+            Assertions::fail,
```
breedx-nr, line 59: Solid, fail immediately if this gets run!  👍

telemetry/src/test/java/com/newrelic/telemetry/LimitingSchedulerTest.java
```diff
@@ -70,7 +68,7 @@ void testScheduleFailsWhenOverCapacity() throws Exception {
   }
 
   @Test
-  void testFirstSuccessSecondFailsRetry() throws Exception {
+  void testSingleOverLimitFails() throws Exception {
```
breedx-nr, line 71: Good change...bad name was bad.

breedx-nr (approved):
Should do a reformat `./gradlew goJF` before merging, but otherwise looks like some good test enhancements.  Thanks.

PR #255, opened 2020-12-02
Telemetryclient notifications

tspring:
Add the notion of a notification handler to capture feedback from the TelemetryClient

### DIFF
--- a/telemetry/src/main/java/com/newrelic/telemetry/LoggingNotificationHandler.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/LoggingNotificationHandler.java
@@ -2,7 +2,12 @@ package com.newrelic.telemetry;
 
 import org.slf4j.Logger;
 
-/** The default NotificationHandler for logging {@link TelemetryClient} responses */
+/**
+ * The default NotificationHandler for logging {@link TelemetryClient} responses.
+ *
+ * <p>Users may choose to override this implementation to provide additional behavior. For example,
+ * applications using the New Relic Java agent could report errors via the NewRelic.noticeError API.
+ */
 public class LoggingNotificationHandler implements NotificationHandler {
 
   private final Logger logger;
--- a/telemetry/src/main/java/com/newrelic/telemetry/LoggingNotificationHandler.java
+++ b/telemetry/src/main/java/com/newrelic/telemetry/LoggingNotificationHandler.java
@@ -1,0 +1,24 @@
+package com.newrelic.telemetry;
+
+import org.slf4j.Logger;
+
+/** The default NotificationHandler for logging {@link TelemetryClient} responses */
+public class LoggingNotificationHandler implements NotificationHandler {
+
+  private final Logger logger;
+
+  LoggingNotificationHandler(Logger logger) {
+    this.logger = logger;
+  }
+
+  @Override
+  public void noticeInfo(
+      String message, Exception exception, TelemetryBatch<? extends Telemetry> batch) {
+    logger.info(message, exception);
+  }
+
+  @Override
+  public void noticeError(String message, Throwable t, TelemetryBatch<? extends Telemetry> batch) {
+    logger.error(message, t);
+  }
+}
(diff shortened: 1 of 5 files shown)


telemetry/src/main/java/com/newrelic/telemetry/NotificationHandler.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param message the message provided by {@link TelemetryClient}
+   * @param batch the telemetry batch that was sent
+   */
+  default void noticeError(String message, TelemetryBatch<? extends Telemetry> batch) {
+    noticeError(message, null, batch);
+  }
+
+  /**
+   * Notice info events
```
aepuls, line 36: This looks like a copy paste error

telemetry/src/main/java/com/newrelic/telemetry/TelemetryClient.java
```diff
@@ -199,7 +200,7 @@ private void scheduleBatchSend(
       scheduler.schedule(
           batch.size(), () -> sendWithErrorHandling(sender, batch, backoff), waitTime, timeUnit);
     } catch (RejectedExecutionException e) {
-      LOG.error("Problem scheduling batch : " + e.getMessage());
+      notificationHandler.noticeError("Problem scheduling batch : ", e, batch);
```
aepuls, line 203: Since the notificationHandler is not final and can be set using the with method, it probably makes sense to put in null checks.

telemetry/src/main/java/com/newrelic/telemetry/LoggingNotificationHandler.java
```diff
@@ -0,0 +1,24 @@
+package com.newrelic.telemetry;
+
+import org.slf4j.Logger;
+
+/** The default NotificationHandler for logging {@link TelemetryClient} responses */
```
aepuls, line 5: You might want to include some information on why you would want to override this. For example, if you are using the New Relic Java agent in addition to the SDK, you could add an implementation to call NewRelic.noticeError.

aepuls:
I added a few small comments.

telemetry/src/test/java/com/newrelic/telemetry/TelemetryClientTest.java
```diff
@@ -156,6 +157,29 @@ void sendGeneratesRetryWithRequestedBackoff() throws Exception {
...
   }
 
+  @Test
+  void sendGeneratesRetryWithRequestedBackoffWithCustomNotificationHandler() throws Exception {
+    CountDownLatch sendLatch = new CountDownLatch(1);
+    when(batchSender.sendBatch(metricBatch))
+        .thenAnswer(
+            invocation -> {
+              throw new RetryWithRequestedWaitException(15, TimeUnit.MILLISECONDS);
+            })
```
jack-berg, line 167: Can be simplified to:
```
        .thenThrow(new RetryWithRequestedWaitException(15, TimeUnit.MILLISECONDS))
```

telemetry/src/test/java/com/newrelic/telemetry/TelemetryClientTest.java
```diff
@@ -219,12 +252,18 @@ void retryWithSplitGoesOverLimit() throws Exception {
...
     testClass.sendBatch(batch);
     testClass.shutdown();
     boolean result = sendLatch.await(3, TimeUnit.SECONDS);
     assertTrue(result);
     assertEquals(1, sendCount.get());
+    assertEquals(0, customNotificationHandler.errorMessages.size());
+    assertEquals(1, customNotificationHandler.infoMessages.size());
+    assertEquals(
+        "Metric batch size too large, splitting and retrying.",
+        customNotificationHandler.infoMessages.get(0));
```
jack-berg, line 266: Consider combining these types of assertions:
```
    assertEquals(Collections.emptyList(), customNotificationHandler.errorMessages);
    assertEquals(Collections.singletonList("Metric batch sending failed. Retrying failed batch after 15 MILLISECONDS"),
            customNotificationHandler.infoMessages);
```
(This looks nicer with Java 11's `List.of()` - we should think about using Java 11+ compatibility for test code)

jack-berg (commented):
Don't know the context of this PR so I won't approve it, but just scrolling through it and noticed a few things that you can take or leave.

jack-berg (approved):

PR #259, opened 2020-12-17
optimize a few code paths to reduce object instantiation

jack-berg:
While investigating how the sdk handles `null` values for `Attributes`, I found a few instances where the code could be optimized.


telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBatchSender.java
```diff
@@ -47,7 +48,7 @@
   public Response sendBatch(EventBatch batch) throws ResponseException {
     if (batch == null || batch.size() == 0) {
       logger.debug("Skipped sending of an empty event batch.");
-      return new Response(202, "Ignored", "Empty batch");
+      return EMPTY_BATCH_RESPONSE;
```
jack-berg, line 51: `Reponse` is an immutable object composed of all immutable fields. No reason to create a new one every time this is called.

telemetry-core/src/main/java/com/newrelic/telemetry/events/json/EventBatchMarshaller.java
```diff
@@ -36,9 +40,44 @@ public String toJson(EventBatch batch) {
             .getTelemetry()
             .stream()
             .map(decorator)
-            .map(new EventToJson())
+            .map(EventBatchMarshaller::mapToJson)
```
jack-berg, line 43: No reason to create a new `EventToJson` mapper instance each time this is invoked. Can refactor `EventToJson` to be a method whose signature matches `Function<String, String>` and use a method reference instead.

telemetry-core/src/main/java/com/newrelic/telemetry/json/AttributesJson.java
```diff
@@ -16,7 +16,6 @@
   public String toJson(Map<String, Object> attributes) {
     StringWriter out = new StringWriter();
     Map<String, Object> filteredAttributes = filterIllegalValues(attributes);
-    filterIllegalValues(filteredAttributes);
```
jack-berg, line 19: `filterIllegalValues` returns a new map with illegal values omitted. Called this here doesn't do anything.

tspring (approved):

telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBatchSender.java
Wald76: Hey, nice catch!  (This is Jim)

Issue #260, opened 2021-01-06
NullPointerException using the telemetry-all dependency


andyduncanitv:
Hi,

I'm seeing a NullPointerException when using the telemetry-all dependency rather than the individual dependencies:

```
0    [main] ERROR com.newrelic.telemetry.transport.BatchDataSender  - Error reading version. Defaulting to 'UnknownVersion'
java.lang.NullPointerException
	at java.base/java.io.Reader.<init>(Reader.java:167)
	at java.base/java.io.InputStreamReader.<init>(InputStreamReader.java:72)
	at com.newrelic.telemetry.transport.BatchDataSender.readVersion(BatchDataSender.java:217)
	at com.newrelic.telemetry.transport.BatchDataSender.<clinit>(BatchDataSender.java:46)
	at com.newrelic.telemetry.events.EventBatchSender.create(EventBatchSender.java:85)
	at com.newrelic.telemetry.events.EventBatchSender$create.call(Unknown Source)
```

This doesn't happen when using the telemetry, telemetry-core, and telemetry-http-java11 separately. I think this is because the telemetry-all packaging doesn't include the [generated version file](https://github.com/newrelic/newrelic-telemetry-sdk-java/blob/main/telemetry-core/build.gradle.kts#L67).

tbradellis:
https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/266

PR #266, opened 2021-01-09
null guard, just print message

tbradellis:
https://github.com/newrelic/newrelic-telemetry-sdk-java/issues/260
This just adds a null guard for the InputStream.
Considered removing the exception from the log all-together, but changed to getMessage() for IOExceptions.


telemetry-core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
```diff
@@ -234,9 +234,13 @@ private static String readVersion() {
...
               .getClassLoader()
               .getResourceAsStream("telemetry.sdk.version.properties");
+
+      if (in == null) {
+        return "UnknownVersion";
+      }
       return new BufferedReader(new InputStreamReader(in)).readLine().trim();
     } catch (Exception e) {
-      logger.error("Error reading version. Defaulting to 'UnknownVersion'", e);
+      logger.error("Error reading version. Defaulting to 'UnknownVersion'", e.getMessage());
```
jack-berg, line 243: This isn't write. When you change the second argument from a throwable to a string you change the overloaded method you're invoking from:
```
    /**
     * Log an exception (throwable) at the ERROR level with an
     * accompanying message.
     *
     * @param msg the message accompanying the exception
     * @param t   the exception (throwable) to log
     */
    public void error(String msg, Throwable t);
```
to:
```
    /**
     * Log a message at the ERROR level according to the specified format
     * and argument.
     * <p/>
     * <p>This form avoids superfluous object creation when the logger
     * is disabled for the ERROR level. </p>
     *
     * @param format the format string
     * @param arg    the argument
     */
     public void error(String format, Object arg)
```
The new argument you're invoking won't use the exception message in any way because there isn't any argument placeholder in the message string template. 

Also, if you're going to add a null check and short circuit return "Unknown Version" so that a NPE isn't generated, we should keep logging the full exception stack trace. It will no longer be triggered for the case where `in == null`, and may help future maintainers track down the source of a problem. 

Also, now that there are two usages of "UnknownVersion", we should create a variable to hold that value. Let's use:
```
private static final String UNKNOWN_VERSION = "UnknownVersion";
...
if (in == null) {
  return UNKNOWN_VERSION;
}
... etc.
```

jack-berg (changes requested):

jack-berg (approved):

PR #268, opened 2021-01-22
ensure dependency module-info.class dont make it into telemetry-core …

jack-berg:
Breaking out part of this larger [PR](https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/267) that fixes the `telemetry-core` shadowJar, which was erroneously including the `module-info.class` file from its `com.google.gson` dependency.


kittylyst (approved):
LGTM. 🚢

PR #271, opened 2021-02-11
Initial cut of adding license key support

kittylyst:


telemetry-core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
```diff
@@ -124,7 +141,11 @@ private Response sendPayload(byte[] payload, UUID requestId, String batchType)
       throws DiscardBatchException, RetryWithSplitException, RetryWithBackoffException,
           RetryWithRequestedWaitException {
     Map<String, String> headers = new HashMap<>();
-    headers.put("Api-Key", apiKey);
+    if (useLicenseKey) {
+      headers.put("X-License-Key", apiKey);
+    } else {
+      headers.put("Api-Key", apiKey);
+    }
```
jack-berg, line 148: You could avoid having to do this if / else every time you send the payload by initializing some field with the name of the api key header at construction time. I.e.:
```
this.apiKeyHeader = useLicenseKey ? "X-License-Key" : "Api-Key";

....
Map<String, String> headers = new HashMap<>();
headers.put(this.apiKeyHeader, apiKey);
```

telemetry-core/src/main/java/com/newrelic/telemetry/SenderConfiguration.java
```diff
@@ -101,6 +118,15 @@ public SenderConfigurationBuilder endpoint(URL endpoint) {
       return this;
     }
 
+    /**
+     * @param useLicenseKey flag to indicate license key should be used instead of insights
+     * @return
```
jack-berg, line 123:
```suggestion
     * @param useLicenseKey flag to indicate the configured {@code apiKey} is a license key, not an insights api key
     * @return this builder
```

jack-berg (approved):
Just a couple minor comments.

PR #272, opened 2021-03-12
update docs with license key

XiXiaPdx:


jasonjkeller:
@XiXiaPdx Can you also update any references to 0.11.0 in the readme to instead be 0.12.0? Once that is done this should be good to merge.

jasonjkeller (approved):

PR #274, opened 2021-06-01
Updated Examples with License Key

yamnihcg:


telemetry_examples/src/main/java/com/newrelic/telemetry/examples/ConfigurationExamples.java
```diff
@@ -73,31 +74,31 @@ public static void main(String[] args) throws MalformedURLException {
                                 .connectTimeout(Duration.ofSeconds(3))
                                 // configure custom stuff here, like proxies, etc.
                                 .build()))
-                .configureWith(insertApiKey)
+                .configureWith(licenseKey)
```
XiXiaPdx, line 77: I think this might be missing `useLicenseKey(true)` ?

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/ConfigurationExamples.java
```diff
@@ -73,31 +74,31 @@ public static void main(String[] args) throws MalformedURLException {
...
                 .endpoint(new URL("http://special-metrics.com/your/custom/path"))
                 .build());
 
     // Configure your span sender:
     SpanBatchSender spanBatchSender =
         SpanBatchSender.create(
             SpanBatchSenderFactory.fromHttpImplementation(Java11HttpPoster::new)
-                .configureWith(insertApiKey)
+                .configureWith(licenseKey)
```
XiXiaPdx, line 85: `useLicenseKey(true)` ? 

Also, for the next two configurations as well.

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
```diff
@@ -35,12 +35,19 @@
...
+    /*
     TelemetryClient telemetryClient =
         TelemetryClient.create(
             () -> new OkHttpPoster(Duration.of(10, ChronoUnit.SECONDS)), insightsInsertKey);
+    */
+
+    // create a TelemetryClient with an http connect timeout of 10 seconds.
+    TelemetryClient telemetryClient =
+        TelemetryClient.create(
+            () -> new OkHttpPoster(Duration.of(10, ChronoUnit.SECONDS)), licenseKey);
```
XiXiaPdx, line 50: I tried this with my prod license key and I'm getting 403. I don't think the code path will allow a license key in this particular example.  Did you get this to work with a license key?

XiXiaPdx (changes requested):

telemetry_examples/src/main/java/com/newrelic/telemetry/examples/TelemetryClientExample.java
XiXiaPdx: @yamnihcg  I don't think a license key will work here, without some changes to the create of the TelemetryClient.  

Lets update the TelemetryClient so that it configured with a license key

XiXiaPdx (approved):
Awesome, thank you! 🚢

PR #277, opened 2021-07-08
Initial commit of IngestWarnings file (checking Ingest API limits) and associated uses in the SDK

yamnihcg:
This PR logs warnings if a user sends data that exceeds the New Relic Ingest API limits. For reference, here are the New Relic Ingest API limits for [Events](https://docs.newrelic.com/docs/telemetry-data-platform/ingest-apis/introduction-event-api/#limits), [Metrics](https://docs.newrelic.com/docs/telemetry-data-platform/ingest-apis/metric-api/metric-api-limits-restricted-attributes/), and [Logs](https://docs.newrelic.com/docs/logs/log-management/log-api/introduction-log-api/#limits). In this PR, we are checking if data exceeds the following limits:

- Maximum number of attributes per Event / Metric / Log 
- Maximum attribute name length 
- Maximum attribute value length 

The IngestWarnings class contains all of the logic that checks if data exceeds Ingest API limits. The corresponding test file, IngestWarningsTest, contains unit tests to verify the functionality of IngestWarnings. 

A user can create an instance of the IngestWarnings class in the SDK. After creating this instance, a user can call the raiseIngestWarnings method to log warnings for a given piece of data. The raiseIngestWarnings method requires 2 arguments. The first argument is the Attributes object associated with the data. The second argument is the data type (Metric, Event, or Log). 

Below is a code snippet from the EventBuffer class that shows how IngestWarnings is used to log warnings in the SDK. 

```
private final IngestWarnings ingestWarnings = new IngestWarnings();

public void addEvent(Event event) {
    Map<String, Object> attributes = event.getAttributes().asMap();
    ingestWarnings.raiseIngestWarnings(attributes, "Event");
    events.add(event);
}
```

### DIFF
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
@@ -8,6 +8,7 @@ import static java.util.Collections.unmodifiableMap;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * A collection of key-value pairs that can be used as the dimensional attributes for metrics.
@@ -86,6 +87,27 @@ public class Attributes {
   }
 
   /**
+   * Remove a attribute and its value.
+   *
+   * @param key to locate the value
+   * @return this
+   */
+  public Attributes remove(String key) {
+    rawAttributes.remove(key);
+    return this;
+  }
+
+  /**
+   * Get the value of an attribute.
+   *
+   * @param key to locate the value
+   * @return this
+   */
+  public Object get(String key) {
+    return rawAttributes.get(key);
+  }
+
+  /**
    * Make a copy of these attributes.
    *
    * @return An unmodifiable copy of these attributes, as a Map.
@@ -97,6 +119,16 @@ public class Attributes {
   /** @return true if there are no attributes in this Attributes instance */
   public boolean isEmpty() {
     return rawAttributes.isEmpty();
+  }
+
+  /** @return number of attributes */
+  public int numAttributes() {
+    return rawAttributes.size();
+  }
+
+  /** @return attribute names */
+  public Set<String> attributeNames() {
+    return rawAttributes.keySet();
   }
 
   @Override
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
@@ -8,7 +8,6 @@ import static java.util.Collections.unmodifiableMap;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 /**
  * A collection of key-value pairs that can be used as the dimensional attributes for metrics.
@@ -87,17 +86,6 @@ public class Attributes {
   }
 
   /**
-   * Remove a attribute and its value.
-   *
-   * @param key to locate the value
-   * @return this
-   */
-  public Attributes remove(String key) {
-    rawAttributes.remove(key);
-    return this;
-  }
-
-  /**
    * Get the value of an attribute.
    *
    * @param key to locate the value
@@ -119,16 +107,6 @@ public class Attributes {
   /** @return true if there are no attributes in this Attributes instance */
   public boolean isEmpty() {
     return rawAttributes.isEmpty();
-  }
-
-  /** @return number of attributes */
-  public int numAttributes() {
-    return rawAttributes.size();
-  }
-
-  /** @return attribute names */
-  public Set<String> attributeNames() {
-    return rawAttributes.keySet();
   }
 
   @Override
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
@@ -86,16 +86,6 @@ public class Attributes {
   }
 
   /**
-   * Get the value of an attribute.
-   *
-   * @param key to locate the value
-   * @return this
-   */
-  public Object get(String key) {
-    return rawAttributes.get(key);
-  }
-
-  /**
    * Make a copy of these attributes.
    *
    * @return An unmodifiable copy of these attributes, as a Map.
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
@@ -41,7 +41,7 @@ public final class EventBuffer {
    */
   public void addEvent(Event event) {
     Map<String, Object> attributes = event.getAttributes().asMap();
-    ingestWarnings.raiseIngestWarnings(attributes, "Event");
+    ingestWarnings.raiseIngestWarnings(attributes, event);
     events.add(event);
   }
 
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
@@ -20,9 +20,6 @@ import org.slf4j.LoggerFactory;
  */
 public final class EventBuffer {
   private static final Logger logger = LoggerFactory.getLogger(EventBuffer.class);
-  private static final int maxNumberOfAttributes = 254;
-  private static final int maxAttributeNameLength = 255;
-  private static final int maxAttributeValueLength = 4096;
   private final Queue<Event> events = new ConcurrentLinkedQueue<>();
   private final IngestWarnings ingestWarnings = new IngestWarnings();
   private final Attributes commonAttributes;
@@ -44,11 +41,7 @@ public final class EventBuffer {
    */
   public void addEvent(Event event) {
     Map<String, Object> attributes = event.getAttributes().asMap();
-
-    ingestWarnings.isValidNumberOfAttributes(attributes, "Event");
-    ingestWarnings.validAttributeNames(attributes);
-    ingestWarnings.validAttributeValues(attributes);
-
+    ingestWarnings.raiseIngestWarnings(attributes, "Event");
     events.add(event);
   }
 
--- a/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
+++ b/telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
@@ -5,10 +5,9 @@
 package com.newrelic.telemetry.events;
 
 import com.newrelic.telemetry.Attributes;
+import com.newrelic.telemetry.util.IngestWarnings;
 import com.newrelic.telemetry.util.Utils;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Queue;
+import java.util.*;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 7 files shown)


telemetry-core/src/main/java/com/newrelic/telemetry/Attributes.java
```diff
@@ -85,6 +86,27 @@ public Attributes put(String key, boolean value) {
     return this;
   }
 
+  /**
+   * Remove a attribute and its value.
+   *
+   * @param key to locate the value
+   * @return this
+   */
+  public Attributes remove(String key) {
```
XiXiaPdx, line 95: Can you explain why `remove` is necessary?

telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
```diff
@@ -21,9 +20,11 @@
  */
 public final class EventBuffer {
   private static final Logger logger = LoggerFactory.getLogger(EventBuffer.class);
-
+  private static final int maxNumberOfAttributes = 254;
```
XiXiaPdx, line 23: Do these need to be here? I thought the goal was to keep it in one place

telemetry-core/src/main/java/com/newrelic/telemetry/events/EventBuffer.java
```diff
@@ -21,9 +20,11 @@
...
+  private static final int maxAttributeNameLength = 255;
+  private static final int maxAttributeValueLength = 4096;
   private final Queue<Event> events = new ConcurrentLinkedQueue<>();
-
+  private final IngestWarnings ingestWarnings = new IngestWarnings();
```
XiXiaPdx, line 27: Can a new  instance  of IngestWarnings be injected through the EventBuffer constructor?

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
```diff
@@ -23,9 +25,9 @@
  */
 public final class MetricBuffer {
   private static final org.slf4j.Logger logger = LoggerFactory.getLogger(MetricBuffer.class);
-
+  private static final int maxMetricAttributes = 100;
```
XiXiaPdx, line 28: CAP case and inject instance into the buffer?

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
```diff
@@ -44,9 +46,40 @@ public MetricBuffer(Attributes commonAttributes) {
    * @param metric The new {@link Metric} instance to be sent.
    */
   public void addMetric(Metric metric) {
+    System.out.println(metric.getClass());
```
XiXiaPdx, line 49: probably remove this

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
```diff
@@ -44,9 +46,40 @@ public MetricBuffer(Attributes commonAttributes) {
...
     metrics.add(metric);
   }
 
+  public void addMetric(Count countMetric) {
+    Map<String, Object> attributes = countMetric.getAttributes();
+
+    ingestWarnings.isValidNumberOfAttributes(attributes, "Metric");
```
XiXiaPdx, line 56: lets refactor the three ingest warnings checks into one method, so less duplication throughout this class

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,88 @@
+package com.newrelic.telemetry.util;
+
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IngestWarnings {
+  private static final Logger logger = LoggerFactory.getLogger(IngestWarnings.class);
+  private static final int maxNumberOfEventAttributes = 254;
```
XiXiaPdx, line 9: cap case and do these match public docs? the 254 especially

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,88 @@
...
+    logger.warn(
+        "The number of attributes in this metric is greater than the maximum allowed attributes per metric.");
+  }
+
+  public void logWarningNumAttributes() {
+    logger.warn(
+        "The number of attributes in this log is greater than the maximum allowed attributes per log.");
+  }
+
+  public void isValidNumberOfAttributes(Map<String, Object> attributes, String dataType) {
```
XiXiaPdx, line 33: Is this used anywhere or just tests?

XiXiaPdx (changes requested):
initial pass

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
```diff
@@ -39,12 +40,36 @@ public MetricBuffer(Attributes commonAttributes) {
   }
 
   /**
-   * Append a {@link Metric} to this buffer, to be sent in the next {@link MetricBatch}.
+   * Append a {@link Count} to this buffer, to be sent in the next {@link MetricBatch}.
+   *
+   * @param countMetric The new {@link Count} instance to be sent.
+   */
+  public void addMetric(Count countMetric) {
```
XiXiaPdx, line 47: Could you help me understand why the three addMetrics with specific parameter types instead of the original one?

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,91 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IngestWarnings {
+  private static final Logger logger = LoggerFactory.getLogger(IngestWarnings.class);
+  private static final int MAX_NUMBER_OF_EVENT_ATTRIBUTES =
+      254; // Public documentation says 255, but true maximum is 254
+  private static final int MAX_NUMBER_OF_METRIC_ATTRIBUTES = 100;
+  private static final int MAX_NUMBER_OF_LOG_ATTRIBUTES =
+      254; // Public documentation says 255, but true maximum is 254
```
XiXiaPdx, line 13: Please add something that would make it possible to find the internal slack conversation where this was discussed. The link to the specific slack comment would be good.

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  public void attributeValueWarning(String attributeValue) {
+    logger.warn(
+        "The value of the attribute "
+            + attributeValue
+            + " is greater than the maximum length allowed for an attribute value.");
+  }
+
+  public void raiseIngestWarnings(Map<String, Object> attributes, String dataType) {
```
XiXiaPdx, line 47: Two things:

Could you move this method to the top of this class please?  :) 

Second, I think with a small tweak, we can make this more flexible.  Events, Metrics, and Logs all implement Telemetry. So we can accept that interface type instead and no longer need to hard code a string.  

You can get the actual type once inside the IngestWarnings and do the logic from there. You might consider saving that type information to be used in string formatting in the log messages too.

```
 public void raiseIngestWarnings(Map<String, Object> attributes, Telemetry dataType) {
     String name = dataType.getClass().getSimpleName();
  }
```

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  public void logWarningNumAttributes() {
+    logger.warn(
+        "The number of attributes in this log is greater than the maximum allowed attributes per log.");
+  }
+
+  public void attributeNameWarning(String attributeName) {
+    logger.warn(
+        "The length of the attribute named "
```
XiXiaPdx, line 35: For these log messages, I would like for you to look into using string formatting. It'll look similar to this pattern in the batchDataSender.  If you use the type information you'll be getting from the class (see comment. below), you can also remove the word log - metric - event from the message and replace with {}.  

```
      logger.debug("Sending json for {} : {} ", batchType, json);
```

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,91 @@
...
+    if (dataType.equals("Metric")) {
+      if (numberOfAttributes > MAX_NUMBER_OF_METRIC_ATTRIBUTES) {
+        metricWarningNumAttributes();
+      }
+    }
+    if (dataType.equals("Log")) {
+      if (numberOfAttributes > MAX_NUMBER_OF_LOG_ATTRIBUTES) {
+        logWarningNumAttributes();
+      }
+    }
```
XiXiaPdx, line 66: Using string formatting, pretty sure you can refactor this down. you might not even need a warning message method. and could be inlined right here :)

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  public void attributeNameWarning(String attributeName) {
+    logger.warn(
+        "The length of the attribute named "
+            + attributeName
+            + " is greater than the maximum length allowed for an attribute name.");
+  }
+
+  public void attributeValueWarning(String attributeValue) {
```
XiXiaPdx, line 40: I think you might be able to refactor atttribute warning into one method with string interpolation and accept another parameter...

XiXiaPdx (changes requested):
It's looking good! Opportunity to reduce the code and make it a bit less fixed,  please let me know if you have any questions.

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
yamnihcg: This was done in anticipation of the batching change (since we need to know the exact event type to generate the JSON).

telemetry-core/src/main/java/com/newrelic/telemetry/util/IngestWarnings.java
```diff
@@ -1,67 +1,44 @@
...
+  public void raiseIngestWarnings(Map<String, Object> attributes, Telemetry dataType) {
 
     // First Check - Check for valid number of attributes
 
     int numberOfAttributes = attributes.size();
-    if (dataType.equals("Event")) {
+    if (dataType instanceof Event) {
       if (numberOfAttributes > MAX_NUMBER_OF_EVENT_ATTRIBUTES) {
-        eventWarningNumAttributes();
+        warningNumAttributes("Event");
```
XiXiaPdx, line 31: Nice, looking better! 

One more tweak, could you try this: 

warningNumAttributes(dataType.getClass().getSimple())

XiXiaPdx (commented):
a quick tweak

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
XiXiaPdx: Right, because each Metric type has a different format

XiXiaPdx (approved):

PR #278, opened 2021-08-06
Add Batching Feature / Configuration to the New Relic Telemetry SDK

yamnihcg:


XiXiaPdx (commented):
@yamnihcg  Lets remove this since it is not being used

telemetry-core/src/main/java/com/newrelic/telemetry/metrics/MetricBuffer.java
```diff
@@ -103,6 +118,106 @@ public MetricBatch createBatch() {
...
+    // Construct JSON for common attributes and add to uncompressed batch size
+
+    String commonJson = generateCommonJSON();
+
+    currentUncompressedBatchSize += commonJson.getBytes(StandardCharsets.UTF_8).length;
+
+    // JSON generation + calculating payload size
+
+    Metric curMetric;
+    while ((curMetric = this.metrics.poll()) != null) {
```
XiXiaPdx, line 149: Ok, it looks like this flows this way and I think there is a problem:

get curMetric  (if this metric gets added to the batch, it will exceed size) 
-- uncompressed size is not over, skip to next if block
-- curMetric is a Count, calculate size and add it to the current uncompressized Size. The current uncompressed Size now exceeds max...but it will be too late by the time this gets checked again?
-- curMetric is added to Batch (The batch now exceeds the size) 
-- Uncompressed is greater than Max...create Batch(which is too big) and add to list.

Did I follow this incorrectly?

telemetry-core/src/main/java/com/newrelic/telemetry/transport/BatchDataSender.java
```diff
@@ -132,7 +132,8 @@ public Response send(String json, TelemetryBatch<? extends Telemetry> batch)
   private byte[] compressJson(String result) throws IOException {
     ByteArrayOutputStream compressedOutput = new ByteArrayOutputStream();
     GZIPOutputStream gzipOutputStream = new GZIPOutputStream(compressedOutput);
-    gzipOutputStream.write(result.getBytes(StandardCharsets.UTF_8));
+    gzipOutputStream.write(
+        result.getBytes(StandardCharsets.UTF_8)); // Used to get number of bytes in piece of data
```
XiXiaPdx, line 136: remove this comment?

XiXiaPdx (changes requested):

XiXiaPdx (changes requested):
This is looking good...please take a look!

